PR #2, opened 2016-03-23
Fixes compatibility with Java 7

chebum:
I fixed several places to make the plugin compatible with JDK 7. I believe it's important since Java 7 is still used in about half of all Java projects (https://plumbr.eu/blog/java/java-version-statistics-2015-edition). We're still working on Java 7 as well.

The most important differences are:
* I replaced **Instant** with **Date**. Since Rollbar expects timestamp in seconds, even Date will suffice.
* Added an implementation for **String.join**
* Replaced **getOrDefault** with **get**. All occurrences of getOrDefault used null as "default" parameter value. This means behavior didn't change if we replace it with a get method call. 
* Fixed JavaDoc in TestThat.java

### DIFF
--- a/src/main/java/com/rollbar/Rollbar.java
+++ b/src/main/java/com/rollbar/Rollbar.java
@@ -7,7 +7,7 @@ import com.rollbar.payload.data.body.Body;
 import com.rollbar.payload.utilities.ArgumentNullException;
 import com.rollbar.payload.utilities.Validate;
 
-import java.time.Instant;
+import java.util.Date;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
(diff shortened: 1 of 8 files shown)


Crisfole:
Hey, you're the man! I'll run the tests (still need to setup CircleCI for this repo), and review the changes.

Eventually we'll probably need to support 6 as well :( but I wasn't entirely sure how to get it all working in different language versions. (I'm a C# guy impersonating a Java programmer...)

Unrelated question: how does the library "feel" to you? I'm still not 100% sold on the completely immutable objects. I feel like I could probably just make the mandatory fields immutable and get a similar feel with a lot less work involved in altering the payloads...

Crisfole:
Tests Pass! Thanks for this. Any feedback is welcome on this!

Crisfole:
Released. Waiting on Maven Central to sync now.

chebum:
> Unrelated question: how does the library "feel" to you? I'm still not 100% sold on the completely immutable objects. I feel like I could probably just make the mandatory fields immutable and get a similar feel with a lot less work involved in altering the payloads...

I don't feel that's important since most of the time you won't interact with it. I created a singleton object as recommended and that was enough.

PR #4, opened 2016-04-13
Adding a test for the null pointer exception I ran across today.

berlin-ab:
Should the constructor take another argument, or should the constructor
be more defensive?

### DIFF
--- a/rollbar/src/main/java/com/rollbar/Rollbar.java
+++ b/rollbar/src/main/java/com/rollbar/Rollbar.java
@@ -83,6 +83,8 @@ public class Rollbar {
                    PayloadFilter filter, PayloadTransform transform) {
         Validate.isNotNullOrWhitespace(accessToken, "accessToken");
         Validate.isNotNullOrWhitespace(environment, "environment");
+
+
         this.sender = sender == null ? new PayloadSender() : sender;
         this.accessToken = accessToken;
         this.environment = environment;
@@ -94,7 +96,7 @@ public class Rollbar {
         this.request = request;
         this.person = person;
         this.server = server;
-        this.custom = new LinkedHashMap<String, Object>(custom);
+        this.custom =  custom == null ? defaultCustom() : wrapCustom(custom);
         this.notifier = notifier == null ? new Notifier() : notifier;
         this.responseHandler = responseHandler;
         this.filter = filter;
@@ -983,5 +985,13 @@ public class Rollbar {
      */
     public Rollbar transform(PayloadTransform transform) {
         return new Rollbar(accessToken, environment, sender, codeVersion, platform, language, framework, context, request, person, server, custom, notifier, responseHandler, filter, transform);
+    }
+
+    private LinkedHashMap<String, Object> wrapCustom(Map<String, Object> custom) {
+        return new LinkedHashMap<String, Object>(custom);
+    }
+
+    private LinkedHashMap<String, Object> defaultCustom() {
+        return new LinkedHashMap<String, Object>();
     }
 }
--- a/rollbar/src/main/java/com/rollbar/Rollbar.java
+++ b/rollbar/src/main/java/com/rollbar/Rollbar.java
@@ -96,11 +96,14 @@ public class Rollbar {
         this.request = request;
         this.person = person;
         this.server = server;
-        this.custom =  custom == null ? defaultCustom() : wrapCustom(custom);
         this.notifier = notifier == null ? new Notifier() : notifier;
         this.responseHandler = responseHandler;
         this.filter = filter;
         this.transform = transform;
+
+        if (custom != null) {
+            this.custom = new LinkedHashMap<String, Object>(custom);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Crisfole:
Hey Adam, was this NPE from having an null in the `custom` field? I have a fix for that in place.

My goal was to have all constructors work as is...I'll go test that locally.

berlin-ab:
Yep, that's the one. We also found another NPE when calling #log.

rollbar/src/test/java/com/rollbar/RollbarTest.java
```diff
@@ -0,0 +1,21 @@
...
+public class RollbarTest {
+
+    @Test
+    public void itDoesNotThrowAnExceptionWhenConfiguringWithJustAccessTokenAndEnvironment() {
+        Boolean exceptionDidHappen = false;
+
+        try {
+            new Rollbar("some-access-token", "some-environment");
+        } catch (NullPointerException e) {
+            exceptionDidHappen = true;
```
Crisfole: You can just use `fail()` here, and then drop the `exceptionDidHappen` field.

Also you might be better off with:

```java
Rollbar r = new Rollbar("Some access token", "environment name");
pass();
```

The uncaught error will propagate up and will display better in the test panels/command line. (Because it will include the exception stack trace).

Crisfole:
@berlin-ab Please do file the other test! This is a new library that I'm really excited about, but it needs some battle hardening.

berlin-ab:
I wanted to make the intention of the test clear and I wasn't sure that raising the exception would do that. I'll try out `fail()` and see how I feel about it.

berlin-ab:
Thanks for making this library.  I've used Rollbar on other projects in different languages, and I was disappointed to not find something more stable for Java.  I'm happy to help.

Crisfole:
@berlin-ab I've also got some `hamcrest` test utilities already present. You can probably find something like 'doesNotThrow' (maybe).

Crisfole:
Better yet, you could probably just create a `testContructorWorks` test:

```java
Rollbar r = new Rollbar("at", "env");
assertEqual("at", r.accessToken());
assertEqual("env", r.environment());
```

berlin-ab:
Ok, updated with production code. 

Side note, I'm not exactly sure what `custom` means for this variable. Can you think of another name that is more descriptive?

rollbar/src/main/java/com/rollbar/Rollbar.java
```diff
@@ -984,4 +986,12 @@ protected void setTransform(PayloadTransform transform) {
     public Rollbar transform(PayloadTransform transform) {
         return new Rollbar(accessToken, environment, sender, codeVersion, platform, language, framework, context, request, person, server, custom, notifier, responseHandler, filter, transform);
     }
+
+    private LinkedHashMap<String, Object> wrapCustom(Map<String, Object> custom) {
+        return new LinkedHashMap<String, Object>(custom);
+    }
+
+    private LinkedHashMap<String, Object> defaultCustom() {
+        return new LinkedHashMap<String, Object>();
```
Crisfole: In this case the 'defaultCustom' would actually need to be null, which would cause it to be skipped in the serializer.

rollbar/src/main/java/com/rollbar/Rollbar.java
```diff
@@ -94,7 +96,7 @@ public Rollbar(String accessToken, String environment, Sender sender, String cod
         this.request = request;
         this.person = person;
         this.server = server;
-        this.custom = new LinkedHashMap<String, Object>(custom);
+        this.custom =  custom == null ? defaultCustom() : wrapCustom(custom);
```
Crisfole: This could probably be more simple:

```java
this.custom = custom == null ? null : new LinkedHashMap<String, Object>(custom);
```

rollbar/src/main/java/com/rollbar/Rollbar.java
```diff
@@ -984,4 +986,12 @@ protected void setTransform(PayloadTransform transform) {
     public Rollbar transform(PayloadTransform transform) {
         return new Rollbar(accessToken, environment, sender, codeVersion, platform, language, framework, context, request, person, server, custom, notifier, responseHandler, filter, transform);
     }
+
+    private LinkedHashMap<String, Object> wrapCustom(Map<String, Object> custom) {
```
Crisfole: This is actually just a copy constructor, not really 'wrapping'.

rollbar/src/test/java/com/rollbar/RollbarTest.java
```diff
@@ -0,0 +1,16 @@
...
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+
+public class RollbarTest {
+
+    @Test
+    public void itDoesNotThrowAnExceptionWhenConfiguringWithJustAccessTokenAndEnvironment() {
```
Crisfole: 👍 Thanks!

Crisfole:
In this case there's literally nothing more descriptive since it contains all the "custom" data you want to send. Literally whatever you want.

Grep for "custom" in here: https://rollbar.com/docs/api/items_post/

berlin-ab:
Ok, I've pushed a new commit that allows custom to be nil and just calls the copy constructor.

Crisfole:
You rock. I'm going to pull this, run the tests, and merge it in, thanks.

Crisfole:
And then I'm going to hook up travis...

berlin-ab:
Great, thanks!

PR #8, opened 2016-04-13
Gradle installation instructions

berlin-ab:
I tend to use Gradle on my projects and appreciate when libraries specify the Gradle installation instructions (even though they are easily translated from maven).

I also did some re-formatting of the README.


Crisfole:
I'm told Gradle's nicer than Maven (I'd believe it...) Thanks for this!

RE: Formatting

Good catch, and sorry about that! I used a really weird markdown parser at my previous job that required all sorts of odd whitespace weirdness. It infected my normal markdown writing.

PR #9, opened 2016-04-13
Null handler test

berlin-ab:
Exposes a null pointer exception when a handler has not been set for the response from the Rollbar server.

I'll clean up the branch history once the constructor null pointer has been merged.


Crisfole:
Merged that last one! You can probably just create a new branch and cherrypick 6e0a090, it doesn't rely on anything that won't be there.

berlin-ab:
@Crisfole I rebased master so only one file has been changed.  I'm not sure how you'd approach solving this problem.

PR #16, opened 2016-05-10
remove version number property, maven cannot inject it during build

dlsteuer:
maven cannot inject the property into the version number and so the project cannot build, I've hardcoded the number, but it can be updated with the maven versions plugin.


coryvirok:
Does this mean we need to update every one of these files whenever we create a new version? Seems like there should be a better way.

Crisfole:
@coryvirok there's a command line way to update all at once that @dlsteuer found. The way this is done currently flat out doesn't work on Maven central for some reason. It'd be simple to create a custom build task that ran that first.

dlsteuer:
@coryvirok @Crisfole I switched it around to using the maven flatten plugin, so we only have to update the version number in the properties under the root pom, but the version number is propagated to all the modules correctly now.  This doesn't require any additional build steps.

PR #28, opened 2016-06-20
Frame json cleanup, and obfuscation fix.

dlsteuer:
If a customer is obfuscating their code, because of the reflection being used for BodyContents, the payload is invalid, this removes the reflection and cleans up the json payload a bit.

### DIFF
--- a/rollbar-payload/src/main/java/com/rollbar/payload/data/body/Frame.java
+++ b/rollbar-payload/src/main/java/com/rollbar/payload/data/body/Frame.java
@@ -217,22 +217,30 @@ public class Frame implements JsonSerializable {
 
     public Map<String, Object> asJson() {
         Map<String, Object> obj = new LinkedHashMap<String, Object>();
-        if (filename() != null)
-            obj.put("filename", filename());
-        if (lineNumber() != null)
-            obj.put("lineno", lineNumber());
-        if (columnNumber() != null)
-            obj.put("colno", columnNumber());
-        if (method() != null)
-            obj.put("method", method());
-        if (code() != null)
-            obj.put("code", code());
-        if (context() != null)
-            obj.put("context", context());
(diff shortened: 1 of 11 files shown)


rollbar-payload/src/main/java/com/rollbar/payload/data/body/Frame.java
```diff
@@ -217,14 +217,22 @@ public Frame keywordArgs(Map<String, Object> keywordArgs) {
...
         Map<String, Object> obj = new LinkedHashMap<String, Object>();
-        obj.put("filename", filename());
-        obj.put("lineno", lineNumber());
-        obj.put("colno", columnNumber());
-        obj.put("method", method());
-        obj.put("code", code());
-        obj.put("context", context());
-        obj.put("args", args());
-        obj.put("kwargs", keywordArgs());
+        if (filename() != null)
```
coryvirok: Can you not call these functions twice? E.g. `filename()`

coryvirok:
Cool.

rollbar-payload/src/main/java/com/rollbar/payload/data/body/Frame.java
```diff
@@ -217,22 +217,30 @@ public Frame keywordArgs(Map<String, Object> keywordArgs) {
...
-            obj.put("method", method());
-        if (code() != null)
-            obj.put("code", code());
-        if (context() != null)
-            obj.put("context", context());
-        if (args() != null)
-            obj.put("args", args());
-        if (keywordArgs() != null)
-            obj.put("kwargs", keywordArgs());
+        String filename = filename();
```
coryvirok: I think this would be more readable if all of the variables were defined at the top of this method and then checked below. Other than this, LGTM.

PR #36, opened 2017-02-12
fixes 35: sets language to 'java' if null

janeklb:


rokob:
Looks good to me, thanks for filing #36 and fixing it.

PR #38, opened 2017-04-19
add support for class_name field

### LINKED ISSUE #34: Incorrect class names in stack traces
Clients based on ahaid/rollbar-logback send "class_name" in the stack trace elements, e.g.:

```
          {
            "class_name": "java.lang.Thread", 
            "method": "run", 
            "lineno": 745, 
            "filename": "Thread.java"
          }, 
          {
            "class_name": "org.eclipse.jetty.util.thread.QueuedThreadPool$3", 
            "method": "run", 
            "lineno": 540, 
            "filename": "QueuedThreadPool.java"
          }, 
          {
            "class_name": "org.eclipse.jetty.util.thread.QueuedThreadPool", 
            "method": "runJob", 
            "lineno": 620, 
            "filename": "QueuedThreadPool.java"
          },
```

which results in a more java-like stacktrace on the UI:

```
128 at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob (QueuedThreadPool.java:620)
129 at org.eclipse.jetty.util.thread.QueuedThreadPool$3.run (QueuedThreadPool.java:540)
130 at java.lang.Thread.run (Thread.java:745)
```

On the other hand this library does not support _class_name_ and sends Frame objects without this field, which is rendered on the UI like this:

```
128 at runJob (org.eclipse.jetty.util.thread.QueuedThreadPool.java:620)
129 at run (org.eclipse.jetty.util.thread.QueuedThreadPool$3.java:540)
130 at main ( java.lang.Thread.java:745)
```

It's a bit odd given that the file names are incorrect (especially that ...$3.java)

Rollbar API spec does not mention _class_name_. 

Could _class_name_ be included (in the spec and in this client) to make stack traces in the UI look like regular java stacktraces - same as the ones from ahaid/rollbar-logback?

mattdrees:
This uses the very-recently-documented API field `class_name` for error reports.
The `filename` field is now filled only with the `StackTraceElement.getFileName()` value, and not the pseudo-filename from before.

I left the old constructor for backwards-compatibility; however I'm not sure it's pulling its weight and doubt many (any?) folks are using it directly. So it might be worth removing. Let me know if you'd like that.

This fixes #34.

### DIFF
--- a/rollbar-payload/src/main/java/com/rollbar/payload/data/body/Frame.java
+++ b/rollbar-payload/src/main/java/com/rollbar/payload/data/body/Frame.java
@@ -18,6 +18,7 @@ public class Frame implements JsonSerializable {
     private final Integer columnNumber;
     private final String method;
     private final String code;
+    private final String className;
     private final CodeContext context;
     private final Object[] args;
     private final LinkedHashMap<String, Object> keywordArgs;
@@ -28,7 +29,7 @@ public class Frame implements JsonSerializable {
      * @throws ArgumentNullException if filename is null
      */
     public Frame(String filename) throws ArgumentNullException {
-        this(filename, null, null, null, null, null, null, null);
+        this(filename, null, null, null, null, null, null, null, null);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


brianr:
@rokob can you review this? Would be great to get this merged so that filename linking works in the Rollbar UI.

mattdrees:
thanks!

PR #46, opened 2017-08-03
Documentation update after moving HTTP to Sender module

xeviknal:
Hello Rollbar team,

I am just new using Rollbar (specially on vanilla Java). I had to integrate you lib with a logger and, therefore, use what you say 'Custom usage'. I saw myself trying to sent a Payload ```p.send()``` unsuccessfully. After some digging into your code I realized that you moved HTTP to sender module. Thus, sending a Payload changed slightly. 

Since documentation wasn't updated accordingly, I decided to send you guys a PR approaching it. I thing this is useful for developers to be updated.


samhatoum:
This threw me off too. Thank you @xeviknal

PR #53, opened 2017-09-22
Add a new set of classes that represents the Rollbar's API and also a…

basoko:
Things I have considered:
* New namespacing for the payloads in the new version with the aim of matching better the current module names. Also,  this will allows us to include the old ones without namespacing collision in case we add the backwards compatibility. 
* All the payloads are immutable and avoid exposing their state by making immutable copies of external collections.
* All internal representation is accessed by accessor methods, in this case getters, like getData.
* Each payload comes with a Builder class to create a fluent api during its construction. Also, this allows us to add properties in a more flexible way without creating constructors to avoid breaking changes. All the payloads have to be build using their proper builder helper. That's why also all the payloads have a private constructor.
* I haven't added any validation to the payloads, this is something we could add or discuss later.
* Added also the google checkstyle to the project to keep the code style in a uniform way.
* Added some tests to the payloads, there are not too many, I just wanted to set up the base of them and at the same time verify things like the `asJson` method that could be critical for json serializaing process in the notifier.

In the `rollbar-api/src/test/java/com/rollbar/test/Factory.java` file you can take a look at how the construction of the payloads looks like.
Also, executing `./gradlew clean build javaDoc` will generate all the reporting regarding the checkstyle, findbugs, tests coverage and documentation, so navigating through the `rollbar-api/build` folder you can find all this information.


rokob:
This looks good to me. I think the change from the old to the new style which is basically going from this:

```
new Request()
        .url("www.google.com")
        .method("GET")
        .headers(headers)
        .get(get)
        .querystring("param1=value1&param2=value2")
        .userIp("192.168.1.1");
```

to this:

```
new Request.Builder()
        .url("https://rollbar.com/project/1")
        .method("GET")
        .headers(headers)
        .get(get)
        .querystring("param1=value1&param2=value2")
        .userIp("192.168.1.1")
        .build();
```

Is not a lot of change in terms of interface, but fits much more naturally with the Java style of doing things and saves on a ton of allocations.

The reporting setup looks great too.

rollbar-api/src/main/java/com/rollbar/api/payload/Payload.java
```diff
@@ -0,0 +1,142 @@
...
+    Payload data = (Payload) o;
+
+    if (accessToken != null ? !accessToken.equals(data.accessToken) : data.accessToken != null) {
+      return false;
+    }
+    return this.data != null ? this.data.equals(data.data) : data.data == null;
+  }
+
+  @Override
+  public int hashCode() {
```
coryvirok: Whenever I see custom `hashCode()` functions I get suspicious. How about we create a temp string `accessToken:data` and then take the `hashcode()` of that string?

rollbar-api/src/main/java/com/rollbar/api/payload/data/Client.java
```diff
@@ -0,0 +1,134 @@
...
+     *
+     * @param clientName the client name.
+     * @param properties the properties of the client.
+     * @return the builder instance
+     */
+    public Builder addClient(String clientName, Map<String, String> properties) {
+      Map<String, String> values = data.get(clientName);
+
+      if (values == null) {
+        values = new HashMap<>();
```
coryvirok: Should this be `HashMap<String, String>()`?

rollbar-api/src/main/java/com/rollbar/api/payload/data/Notifier.java
```diff
@@ -0,0 +1,140 @@
...
+    if (name != null ? !name.equals(notifier.name) : notifier.name != null) {
+      return false;
+    }
+    return version != null ? version.equals(notifier.version) : notifier.version == null;
+  }
+
+  @Override
+  public int hashCode() {
+    int result = name != null ? name.hashCode() : 0;
+    result = 31 * result + (version != null ? version.hashCode() : 0);
```
coryvirok: Same comment as above

rollbar-api/src/main/java/com/rollbar/api/payload/Payload.java
rokob: This is how almost every hashCode implementation I've ever seen does it in Java because the built-in string uses something like this pattern with powers of 31 and the sum of individual parts.

rollbar-api/src/main/java/com/rollbar/api/payload/data/Client.java
basoko: Not needed since java7 when it was introduce the diamond operator to make it less verbose.

rollbar-api/src/main/java/com/rollbar/api/payload/Payload.java
basoko: All hashCode methods have been autogenerated by the IDE using the "facto-standard" way to do it.

rollbar-api/src/main/java/com/rollbar/api/payload/data/Client.java
coryvirok: :) Thanks. It's been a while since I 💻 in ☕️ .

rollbar-api/src/main/java/com/rollbar/api/payload/Payload.java
coryvirok: 👍 Good enough for me

PR #54, opened 2017-09-26
Add rollbar-java definitions.

basoko:
This PR includes:
* Compilation target java7.
* Rollbar notifier with interfaces to retrieve the data.
* A project example to show its usage.

### DIFF
--- a/examples/rollbar-java/src/main/java/com/rollbar/example/Application.java
+++ b/examples/rollbar-java/src/main/java/com/rollbar/example/Application.java
@@ -1,0 +1,98 @@
+package com.rollbar.example;
+
+import static com.rollbar.notifier.config.ConfigBuilder.withAccessToken;
+
+import com.rollbar.api.payload.Payload;
+import com.rollbar.notifier.Rollbar;
+import com.rollbar.notifier.config.Config;
+import com.rollbar.notifier.provider.Provider;
+import com.rollbar.notifier.sender.Result;
+import com.rollbar.notifier.sender.Sender;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+/**
+ * Application example using rollbar-java notifier.
+ */
+public class Application {
+
+  private Greeting greeting;
+
+  private Rollbar rollbar;
+
+  /**
+   * Constructor.
+   */
+  public Application() {
+    this.greeting = new Greeting();
+    Config config = withAccessToken(System.getenv("ROLLBAR_ACCESSTOKEN"))
+        .environment("development")
+        .codeVersion("1.0.0")
+        .language("java")
+        .build();
+    this.rollbar = Rollbar.init(config);
+    this.rollbar.handleUncaughtErrors();
+  }
+
+  /**
+   * Main method.
+   * @param args the args.
+   */
+  public static void main(String[] args) {
+    Application app = new Application();
+    app.execute();
+
+    throw new RuntimeException("Exception finishing execution.");
+  }
+
+  private void execute() {
+    ExecutorService executor = Executors.newFixedThreadPool(3);
+    List<Callable<Void>> callableTasks = new ArrayList<>();
+
+    for (int i = 0; i < 20; i++) {
+      callableTasks.add(new GreetingTask(i + 1, rollbar, greeting));
+    }
+
+    try {
+      executor.invokeAll(callableTasks);
+    } catch (InterruptedException e) {
+      rollbar.error(e, "Unexpectedly the execution was interrupted.");
+    } finally {
+      executor.shutdown();
+    }
+  }
+
+  private static final class GreetingTask implements Callable<Void> {
+
+    private final int taskNumber;
+
+    private final Rollbar rollbar;
+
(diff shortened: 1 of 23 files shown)


rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
```diff
@@ -1,4 +1,633 @@
...
+  public void handleUncaughtErrors() {
+    handleUncaughtErrors(Thread.currentThread());
+  }
+
+  /**
+   * Handle all uncaught errors on {@code thread} with this `Rollbar`.
+   *
+   * @param thread the thread to handle errors on.
+   */
+  public void handleUncaughtErrors(Thread thread) {
```
rokob: I forgot we had this method, it makes sense to then not have `handleUncaughtErrors` as part of the configuration but instead keep it as an instance method. Well, we could have both, we could have it automatically handle errors for the current thread by default / by configuration, and still provide this method that one can also call if they are managing their own threads.

rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
```diff
@@ -1,4 +1,633 @@
...
+  }
+
+  /**
+   * Get the level of the error or message. By default: CRITICAL for {@link Error}, ERROR for other
+   * {@link Throwable}, WARNING for messages. Override to change this default.
+   *
+   * @param error the error.
+   * @return the level.
+   */
+  public Level level(Throwable error) {
```
rokob: This appears to be unused

rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
```diff
@@ -1,4 +1,633 @@
...
+  private void process(Throwable error, Map<String, Object> custom, String description,
+      Level level) {
+
+    // Pre filter
+    if (config.filter() != null && config.filter().preProccess(level, error, custom, description)) {
+      return;
+    }
+
+    // Gather information to build a payload.
+    Data data = buildData(error, custom, description, level);
```
rokob: It took me a second to figure out why this returns the data instead of the builder, and it is because the later transforms take in the whole data object in order to work.

rollbar-java/src/main/java/com/rollbar/notifier/config/ConfigBuilder.java
```diff
@@ -0,0 +1,426 @@
...
+import com.rollbar.notifier.sender.Sender;
+import com.rollbar.notifier.transformer.Transformer;
+import com.rollbar.notifier.uuid.UuidGenerator;
+import java.util.Map;
+
+/**
+ * Configuration builder to build the {@link Config configuration} of the {@link Rollbar rollbar}
+ * notifier.
+ */
+public class ConfigBuilder {
```
rokob: 👍

rollbar-java/src/main/java/com/rollbar/notifier/util/BodyFactory.java
```diff
@@ -0,0 +1,85 @@
...
+      result.add(frame(elements[i]));
+    }
+
+    return result;
+  }
+
+  private static ExceptionInfo info(Throwable throwable, String description) {
+    String className = throwable.getClass().getSimpleName();
+    String message = throwable.getMessage();
+    return new ExceptionInfo.Builder()
```
rokob: These API builders look like they make these much easier to work with.

examples/rollbar-java/src/main/java/com/rollbar/example/Application.java
```diff
@@ -0,0 +1,98 @@
...
+   */
+  public Application() {
+    this.greeting = new Greeting();
+    Config config = withAccessToken(System.getenv("ROLLBAR_ACCESSTOKEN"))
+        .environment("development")
+        .codeVersion("1.0.0")
+        .language("java")
+        .build();
+    this.rollbar = Rollbar.init(config);
+    this.rollbar.handleUncaughtErrors();
```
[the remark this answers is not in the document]
basoko: I just keep this method to keep the API as similar as we had in the old notifier. 

Not sure if I got your idea, but I think that we could have in the `Config` interface a method like this `boolean handleUncaughtErrors()` and then in the Rollbar constructor we could check this flag to make the call to the notifier's handleUncaughtErrors that already sets the handler, is that what you mean?

rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
basoko: Yes, I think we could have both and introduce the Config flag to auto set this param.

Some pending changes when we finish the implementation are related to the default values in the `ConfigBuilder`, where in the case of the `handleUncaughtErrors` if we finally introduce it we could return `true` by default.

rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
basoko: That's right! I forgot to append the level information when building the data. I will commit a change. Thanks!

rollbar-java/src/main/java/com/rollbar/notifier/util/BodyFactory.java
basoko: :smiley:

PR #55, opened 2017-09-27
Add web integration and example.

basoko:
Add the following features:
* Filter to catch all exceptions.
* Servlet request listener to track the current request.
* Providers to extract data from the request. Now the Rollbar request and person should be filled.
* Example of a project using rollbar-web.

Pending things:
* Improve the extraction of the Rollbar request data to get more information, body, post params...
* Simplify the integration using servlet 3.0 features and register the listener and filter automatically through a ServletContainerInitializer.


rollbar-web/src/main/java/com/rollbar/web/listener/RollbarRequestListener.java
```diff
@@ -0,0 +1,26 @@
+package com.rollbar.web.listener;
+
+import javax.servlet.ServletRequestEvent;
+import javax.servlet.ServletRequestListener;
+import javax.servlet.http.HttpServletRequest;
+
+public class RollbarRequestListener implements ServletRequestListener {
+
+  private static final ThreadLocal<HttpServletRequest> CURRENT_REQUEST = new ThreadLocal<>();
```
coryvirok: Are we guaranteed to have a single request per thread? If not, another way of implementing this which I think I prefer, is to attach a new instance of the Rollbar notifier to each request and expose it to the user somehow.

rollbar-web/src/main/java/com/rollbar/web/provider/RequestProvider.java
```diff
@@ -0,0 +1,103 @@
...
+
+        if (values != null && values.length > 0) {
+          StringBuilder sb = new StringBuilder();
+          for (int i = 0; i < values.length; i++) {
+            sb.append(values[i]);
+            if (i + 1 != values.length) {
+              sb.append(',');
+            }
+          }
+          params.put(paramName, sb.toString());
```
coryvirok: Instead of having the params be a string, would it make more sense to have them be a list/array? Otherwise, we wouldn't be able to determine the different between these two query params/values:

```
?param=hello&param=world
?param=hello,world
```

rollbar-web/src/main/java/com/rollbar/web/provider/RequestProvider.java
```diff
@@ -0,0 +1,103 @@
...
+
+    return null;
+  }
+
+  private static String queryString(HttpServletRequest request) {
+    return request.getQueryString();
+  }
+
+  private static String userIp(HttpServletRequest request) {
+    String remoteAddr = request.getHeader("X-FORWARDED-FOR");
```
coryvirok: I think we will want to make this configurable. Not all applications are going to want to use this header to determine the user-ip. E.g. some would use `X-RealIp`.

rollbar-web/src/main/java/com/rollbar/web/filter/RollbarFilter.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+
+public class RollbarFilter implements Filter {
+
+  private Rollbar rollbar;
```
coryvirok: Does this need to be a `ThreadLocal`? Is a filter instantiated for each thread or request?

rollbar-web/src/main/java/com/rollbar/web/filter/RollbarFilter.java
```diff
@@ -0,0 +1,58 @@
...
+  RollbarFilter(Rollbar rollbar) {
+    this.rollbar = rollbar;
+  }
+
+  @Override
+  public void init(FilterConfig filterConfig) throws ServletException {
+    String accessToken = filterConfig.getInitParameter("access_token");
+
+    rollbar = Rollbar.init(withAccessToken(accessToken)
+        .language("java")
```
coryvirok: This should be the default and shouldn't need to be specified.

rollbar-web/src/main/java/com/rollbar/web/provider/RequestProvider.java
```diff
@@ -0,0 +1,103 @@
...
+  public Request provide() {
+    HttpServletRequest req = RollbarRequestListener.getServletRequest();
+
+    if(req != null) {
+      Request request = new Request.Builder()
+          .url(url(req))
+          .method(method(req))
+          .headers(headers(req))
+          .get(getParams(req))
+          .querystring(queryString(req))
```
coryvirok: Can we rename this to be `queryString(...)`?

rollbar-web/src/main/java/com/rollbar/web/filter/RollbarFilter.java
basoko: Sure, this will be set by default by the `ConfigBuilder` located at the module `rollbar-java`, but as at this moment it has not default values assigned I just did it here for testing purposes.

rollbar-web/src/main/java/com/rollbar/web/provider/RequestProvider.java
basoko: Great point! We will need to figure out how to set up this setting.

rollbar-web/src/main/java/com/rollbar/web/provider/RequestProvider.java
basoko: Sure, I'll rename it.

rollbar-web/src/main/java/com/rollbar/web/provider/RequestProvider.java
basoko: Yes, you are right. I'll make the change to the `Request` class located under `rollbar-api` to allow a `Map<String, String[]>` or `Map<String, List<String>>`. It was just done in this way because the previous version was set as `Map<String, String>` but I think it makes more sense to be an array or a list.

rollbar-web/src/main/java/com/rollbar/web/listener/RollbarRequestListener.java
basoko: Yes, AFAIK every request in a servlet container is manage by a different thread. So, if I am not wrong in a servlet container when a request reaches it, it will assign a thread from its "requests" thread pool to handle it.

rollbar-web/src/main/java/com/rollbar/web/filter/RollbarFilter.java
basoko: AFAIK filters and servlets are one instance per application context, that's it per application deployed in the servlet container. 
At the same time, in this implementation we are using the library manage instance so the same instance will be used by every thread. The `rollbar` notifier should be thread-safe, thing that at this point I think it is if the `Providers` are. This should be worth mentioning in the `Provider` interface.

rollbar-web/src/main/java/com/rollbar/web/filter/RollbarFilter.java
```diff
@@ -0,0 +1,71 @@
...
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+
+public class RollbarFilter implements Filter {
+
+  static final String ACCESS_TOKEN_PARAM_NAME = "access_token";
+
+  static final String USER_IP_HEADER_PARAM_NAME = "user_ip_header";
+
+  private Rollbar rollbar;
```
rokob: From my googling around, Filters are reused across threads, so this instance variable will be used across threads and therefore this code is not thread safe unless the Rollbar class is.

basoko: Yes, there is just one instance of a filter per application context. The point is, as I mentioned, that the Rollbar instance should be thread safe, and so far I think it is if the providers are thread safe.

PR #56, opened 2017-09-28
Implementation of minimal functionality

basoko:
This PR includes de following changes:
* Default configuration in the ConfigBuilder. Things like language, notifier provider, etc are pre-set by it.
* Sender implementation based on the previous version. It has been also resuse the json serializer with a minimal refactor.
* Notifier Provider. Now, the notifier information is given by this provider and pre-set as default in the ConfigBuilder.
* TransformerPipeline and FilterPipeline. This utility classes have the purpose make easier to work with a set of transfomers and filters. Both could be used internally by the notifier to set our custom transformations or filters and at the same time for third party users.
* Add configuration flag for auto set the handler for uncaught errors.


rollbar-java/src/main/java/com/rollbar/notifier/sender/json/JsonSerializerImpl.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Implementation of the {@link JsonSerializer json serializer}.
+ */
+public class JsonSerializerImpl implements JsonSerializer {
```
rokob: This looks like the code from the JSON serializer from the current rollbar-java, so I am assuming it is from there. I have no idea why that library had a handrolled JSON serializer, but it seems crazy to me. Can we use Jackson or GSON or something standard?

rollbar-java/src/main/java/com/rollbar/notifier/sender/json/JsonSerializerImpl.java
```diff
@@ -0,0 +1,191 @@
...
+        Map.Entry thisOne = (Map.Entry) o;
+        Object key = thisOne.getKey();
+        Object val = thisOne.getValue();
+        obj.put(key.toString(), val);
+      }
+      return obj;
+    }
+  }
+
+  private static void serializeString(StringBuilder builder, String str) {
```
rokob: There was a bug a fixed in this code in the old library with this function. This doesn't work correctly. So if we do want to use this current implementation then we need to at least fix this to match what rollbar-java currently has.

rollbar-java/src/main/java/com/rollbar/notifier/sender/SyncSender.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  /**
+   * Constructor.
+   *
+   * @param url the Rollbar API endpoint.
+   * @param serializer the serializer.
+   */
+  public SyncSender(String url, JsonSerializer serializer) {
+    try {
```
coryvirok: Should this call the constructor below?

rollbar-java/src/main/java/com/rollbar/notifier/sender/SyncSender.java
```diff
@@ -0,0 +1,136 @@
...
+    HttpURLConnection connection = getConnection();
+    byte[] bytes = body.getBytes(UTF_8);
+    sendJson(connection, bytes);
+    return readResponse(connection);
+  }
+
+  private HttpURLConnection getConnection() throws IOException {
+    HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+
+    connection.setRequestProperty("Accept-Charset", UTF_8);
```
coryvirok: We also need to add the Rollbar access token to the list of headers. E.g. https://github.com/rollbar/rollbar.js/commit/34879c050c4e2e596bca02bd4984448a6836ff6b#diff-6eb48d11db7fa35f913611ed779825adR69

rollbar-java/src/main/java/com/rollbar/notifier/sender/json/JsonSerializer.java
```diff
@@ -0,0 +1,28 @@
...
+  String toJson(Payload payload);
+
+  /**
+   * Parses the response from Rollbar to a {@link Result result}.
+   *
+   * @param code the response code.
+   * @param response the response.
+   * @return the result.
+   */
+  Result resultFrom(int code, String response);
```
coryvirok: Why is the response code needed here?

rollbar-java/src/main/java/com/rollbar/notifier/sender/result/ResultCode.java
```diff
@@ -1,4 +1,4 @@
-package com.rollbar.notifier.sender;
+package com.rollbar.notifier.sender.result;
 
 /**
  * Represents expected response codes from POSTing an item to Rollbar.
```
coryvirok: Do we really need a class for this? I would imagine there's already something in the standard library for mapping the response codes to errors/messages.

rollbar-java/src/main/java/com/rollbar/notifier/sender/json/JsonSerializerImpl.java
basoko: Yes, the code from the JsonSerializer and SyncSender it's coming from the previous version. 
I guess that we were using a handrolled JSON serializer to not introduce a third party dependency in our code that would end up being a transitive dependency for our users, but this is something that can be changed.

rollbar-java/src/main/java/com/rollbar/notifier/sender/json/JsonSerializerImpl.java
basoko: I'll take a look at the last version then and update this code. Thanks for the heads up!

rollbar-java/src/main/java/com/rollbar/notifier/sender/json/JsonSerializer.java
basoko: It's used because in the JsonSerializerImpl the response is parsed through a RegEx to extract the uuid or the error message returned by the API. At this point the RegEx to use is based on this result code, but we could just try one or the other when parsing the response and remove the result code.

rollbar-java/src/main/java/com/rollbar/notifier/sender/SyncSender.java
basoko: Sure, I'll add it.

rollbar-java/src/main/java/com/rollbar/notifier/sender/SyncSender.java
basoko: I'll take a look at it because the approach we are following is based on having builders to not call directly the constructors.

rollbar-java/src/main/java/com/rollbar/notifier/sender/result/ResultCode.java
basoko: As the SyncSender and JsonSerializer this comes from the previous version. We could use instead of it just int numbers mapped by the constants in the [HttpURLConnection](http://docs.oracle.com/javase/7/docs/api/java/net/HttpURLConnection.html). However some codes are missing, like the 429 mapped to `Too many requests`. At the same time this could work as a quick view/doc of the response codes our API delivers.

PR #63, opened 2017-10-05
Android on top of rollbar-java

rokob:
So my reading of rollbar-android is that it is actually not thread safe right now. For example, you can call `Rollbar.setPerson` from different threads and actually change data that is inside a payload that appears to only be living on a single executor thread. Therefore, there is a bit here that is not thread safe that we will need to talk about. Mainly around setting a couple configuration settings, in particular the current person.

I got rid of most of the configuration setters that currently exist in rollbar-android in favor of running everything through the initial configuration object, but we might have to include some locking around setting the person because that is something people should need to do while an app is running, i.e. login/logout. Otherwise, I'm still working on my tests so I didn't include those yet.


basoko:
Nice to see the android integration based on the new notifier Api! Thanks @rokob!

Regarding the `Rollbar.setPerson` method, yes this seems to be not thread safe at this point because some thread could be sending payloads meanwhile other is modifying that person information.

I think that the comomn use case will be many threads that are `reading` the person information through that provider to send payloads to Rollbar, meanwhile suddenly another thread will want to modify it, like the use case you mentioned, login/logout example.

So, for this use case where we would have probably many threads reading very frequently, and very few writing not as frequently, I would suggest to use a ReadWriteLock. With the use of it we could make it thread safe without locking all the threads that wants to send payloads and there is not changes in the configuration.

Also, from my point of view I'd rather not have that `Rollbar.setPerson` that just modifies the `PersonProvider` content, and as it can happen with other providers, we could add a `reload(Config config)` method in the notifier, that would replace the config instance inside the Rollbar notifier. To do this we could use the approach I mentioned above, because the use case would be the same, tons of threads trying to send payloads, and sometimes ones that want to change the conf, so I'd synchronize it with a ReadWriteLock. If you want I could create a PR with this change and we can test this approach.

Regarding the changes I got some small issues trying to build it. Some of them were due to the linter and min version that is seems is set to 1. I did this change to solve them in the `build.gradle` of the rollbar-android.

```
 android {                                                                                                                                                                                                        
     compileSdkVersion 23                                                                                                                                                                                         
     buildToolsVersion '25.0.3'                                                                                                                                                                                   
+                                                                                                                                                                                                                 
+    defaultConfig {                                                                                                                                                                                              
+        minSdkVersion 19                                                                                                                                                                                         
+    }                                                                                                                                                                                                            
 }  
```
Also, in the `rollbar-api/src/test/java/com/rollbar/test/Factory.java` I got some compilation errors that I think are related with the changes done in some types under `rollbar-api`.
Finally, I don't know if it'd be possible to use the same code style for the Android source code and that the one used in the other modules, and how to make a release of an Android library and if we could use the same plugin than we will use for the other modules.

basoko:
One worth thing to note here is that the `BufferedSender` as its based on the `DiskQueue` should just send payloads that contain `Serializable` objects or otherwise it will failed when writing the payloads to disk.

settings.gradle
```diff
@@ -1,8 +1,8 @@
...
+rootProject.name="rollbar-java-sdk"
 
 include ":rollbar-api",
         ":rollbar-java",
         ":rollbar-android",
-        "rollbar-web",
-        ":examples:rollbar-java",
-        ":examples:rollbar-web"
+        "rollbar-web"//,
+        //":examples:rollbar-java",
```
basoko: Should we uncomment this?

rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
```diff
@@ -92,19 +92,37 @@ public void handleUncaughtErrors(Thread thread) {
...
-      this.config = configProvider.provide(builder);
+      builder = ConfigBuilder.withConfig(this.config);
     } finally {
-      this.configWriteLock.unlock();
+      this.configReadLock.unlock();
     }
+
+    Config newConfig = configProvider.provide(builder);
+
+    this.configWriteLock.lock();
```
basoko: I think here we could use the `configure(Config config)` method. 
Also, I'd wrap all the operations after acquiring the lock in a `try` statement, and the release of them in the `finally` clause, although the code is more verbose is a kind of pattern that you always see and could avoid some unexpected behaviors if weird errors happen.

rollbar-api/src/main/java/com/rollbar/api/payload/data/body/Message.java
```diff
@@ -29,7 +32,9 @@ public String getKeyName() {
 
   @Override
   public Object asJson() {
-    return this.body;
+    Map<String, String> message = new HashMap<>();
```
basoko: WoW! Good catch! I didn't test using messages! Thanks!

rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
```diff
@@ -80,6 +89,43 @@ public void handleUncaughtErrors(Thread thread) {
...
+    this.configReadLock.lock();
+    try {
+      builder = ConfigBuilder.withConfig(this.config);
+    } finally {
+      this.configReadLock.unlock();
+    }
+
+    Config newConfig = configProvider.provide(builder);
+
+    this.configWriteLock.lock();
```
basoko: I think we could use here the `configure(Config config)` method.
Also, I'd wrap all the operations after getting the lock in a `try` clause and the release of it in the `finally` clause. This could avoid some weird errors that could happen and its a kind of pattern that you always find, so makes easier to read the code, although it's more verbose.

rokob:
When it comes to style like indentation I didn't change the default settings in IntelliJ to match what you were using so I am pretty sure I screwed some things up. I was planning on going back and fixing it up at the end. What are the style settings you are using?

basoko:
I can push them in the `tools` folder to make them available if you want.

rollbar-java/src/main/java/com/rollbar/notifier/sender/queue/DiskQueue.java
```diff
@@ -37,7 +37,7 @@ private DiskQueue(Builder builder) {
     this.queueFolder = builder.queueFolder;
 
     if (!this.queueFolder.exists()) {
-      if (this.queueFolder.mkdirs()) {
+      if (!this.queueFolder.mkdirs()) {
```
rokob: This was a doozy

PR #65, opened 2017-10-11
Add publishing with maven-publish plugin and signing

basoko:
I think I managed to setup the `maven-publish` plugin with `signing` getting rid of the `maven` one. 

I've created a `gradle/release.gradle` where I moved all this logic and also removed all those `gradle.properties` files.

Now if you execute `./gradlew publishToMavenLocal` you could see how everything will be layout in the repo, and for uploading the archives I think that you should execute `./gradlew publish` plugin if you have set up all the things needed like credentials of the repository and the ones for signing.

@rokob Have in mind that this not publish the `rollbar-android`, maybe another way to publish it is required.

Test it before merge because maybe I broke something regarding the `rollbar-android`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,33 +1,8 @@
-//plugins {
-//    id 'digital.wup.android-maven-publish' version '3.0.0' apply false
-//}
-
-def isReleaseBuild() {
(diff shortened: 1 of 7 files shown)


rokob:
👍

PR #69, opened 2017-10-20
Backwards compatibility stuff

rokob:
Update documentation and add some roughly backwards compatible, but deprecated constructors

### DIFF
--- a/rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
+++ b/rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
@@ -1,15 +1,21 @@
 package com.rollbar.notifier;
 
 import com.rollbar.api.payload.Payload;
-import com.rollbar.api.payload.data.Data;
-import com.rollbar.api.payload.data.Level;
+import com.rollbar.api.payload.data.*;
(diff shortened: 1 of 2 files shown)


rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
```diff
@@ -693,4 +699,103 @@ private void sendPayload(Config config, Payload payload) {
...
   }
+
+  /**
+   * Construct a notifier defaults for everything including Sender.
+   * Caution: default sender is slow and blocking. Consider providing a Sender overload.
+   * @param accessToken not nullable, the access token to send payloads to
+   * @param environment not nullable, the environment to send payloads under
+   */
+  @Deprecated
+  public Rollbar(String accessToken, String environment) {
```
basoko: We are adding the constructors from the previous version but the packaging has changed so it will imply to update the user's code to use the new packaging. 
We could add a class with the same previous API at `com.rollbar` level and implement the old API in it, using delegation to this class. We could also mark as `depecreated` this whole class.

rokob:
I'm going to merge this then move it to the com.rollbar namespace to resolve the packaging in a separate PR that I've got

PR #75, opened 2017-10-30
Not generate war artifact for rollbar-web

basoko:


rokob:
Looks good, I'm going to merge because and test along with some minor android changes I am going to make locally.

PR #77, opened 2017-11-01
Add sl4j and logging traces

### LINKED ISSUE #76: Improve troubleshooting for SSL cert issues
The problem @helderribeirosousa had in #72 is going to come up again for more users. Let's figure out some ways to make this issue, and other issues that manifest just as "it doesn't work", easier to debug and fix.

For the SSL cert issue in particular, we should add some documentation.

basoko:
Fixes #76

### DIFF
--- a/rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
+++ b/rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
@@ -1,31 +1,29 @@
 package com.rollbar.notifier;
 
 import com.rollbar.api.payload.Payload;
-import com.rollbar.api.payload.data.*;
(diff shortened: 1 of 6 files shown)


rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
```diff
@@ -88,7 +87,7 @@ public void handleUncaughtErrors() {
    */
   public void handleUncaughtErrors(Thread thread) {
     Objects.requireNonNull(thread, "thread");
-
+    LOGGER.debug("Handling uncaught errors fot thread: {}.", thread);
```
rokob: Handling uncaught errors fot thread => Handling uncaught errors for thread

PR #84, opened 2017-11-15
Integration test

basoko:


rokob:
The rest of the issues are false positives.

PR #102, opened 2018-02-09
Support whether data comes from an uncaught exception or not

### LINKED ISSUE #93: Add ability to detect if exception if caught or uncaught
and then be able to edit the payload

rokob:
There are certain use cases where people want to know if a set of data
comes from an uncaught exception or not. Add this information to the
data object and to the filter interface. This way we can use our
existing filter and transform infrastructure for dealing with this data.

Fixes #93

### DIFF
--- a/rollbar-java/src/main/java/com/rollbar/notifier/filter/Filter.java
+++ b/rollbar-java/src/main/java/com/rollbar/notifier/filter/Filter.java
@@ -17,11 +17,9 @@ public interface Filter {
    * @param error the error.
    * @param custom the custom data.
    * @param description the description.
-   * @param isUncaught whether or not this set of data originates from an uncaught exception.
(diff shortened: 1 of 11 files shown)


rollbar-java/src/main/java/com/rollbar/notifier/filter/Filter.java
```diff
@@ -21,6 +21,27 @@
    */
```
basoko: I think we can mark this method as deprecated.

rokob: I was trying to change this interface so that this change would not be a breaking change by deprecating this and providing a default implementation of the method below. But it looks like default interface implementations are not supported before Java 8, so the only thing I can see to do is to make a semver major breaking change.

PR #104, opened 2018-02-15
Allow arbitrary extra data in API objects that support this loose format

### LINKED ISSUE #86: support additional attributes for various objects in Payload
The [API docs](https://rollbar.com/docs/api/items_post/) are pretty clear that it's ok to send non-standard attributes in the item JSON, but rollbar-java 1.0.0 has removed this capability. (It could be done for some, but not all, objects in 0.5.4.)

Would you consider re-adding this capability?
I find it useful to record things like the response code, the timestamps at the beginning and the end of the request, extra person attributes, etc.

I'm sorry I wasn't able to get this feedback to you before 1.0.0 was released!

rokob:
In the [API docs](https://rollbar.com/docs/api/items_post/) some of the elements in the payload body support arbitrary extra data with a comment that starts like `// Can contain any arbitrary keys.`. This PR adds back the ability to include this extra data with the various objects that the API understands, namely: `Server`, `Message`, `Request`. `Client` already supports this. Additionally, the docs don't mention it but we have always supported arbitrary extra data with `Person` so this adds that capacity as well.

The general form is to use the Builder for each type and to pass in a `Map<String, Object>` to the `metadata` function. This map will get merged into the final JSON for that object before the standard keys are added. That means that keys that are standard such as `body` for the `Message` type will be overwritten if they exist in this metadata object. See the tests added herein for examples of this behaviour.

Fixes #86

### DIFF
--- a/rollbar-api/src/main/java/com/rollbar/api/payload/data/Person.java
+++ b/rollbar-api/src/main/java/com/rollbar/api/payload/data/Person.java
@@ -22,7 +22,7 @@ public class Person implements JsonSerializable {
     this.id = builder.id;
     this.username = builder.username;
     this.email = builder.email;
-    this.metadata = builder.metadata != null ? unmodifiableMap(builder.metadata) : null;
+    this.metadata = builder.metadata != null ? unmodifiableMap(new HashMap<>(builder.metadata)) : null;
   }
 
   /**
--- a/rollbar-api/src/main/java/com/rollbar/api/payload/data/Person.java
+++ b/rollbar-api/src/main/java/com/rollbar/api/payload/data/Person.java
@@ -1,5 +1,6 @@
 package com.rollbar.api.payload.data;
 
+import static java.util.Collections.unmodifiableMap;
 import com.rollbar.api.json.JsonSerializable;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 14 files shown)


rokob:
This is so weird. The test that is failing on the push should ALWAYS be failing but it passes because the JVM is pretty fast. We are comparing two different calls to the current system time for equality, so I'll fix that.

rollbar-api/src/test/java/com/rollbar/test/Factory.java
```diff
@@ -25,6 +25,8 @@
 
 public class Factory {
 
+  private static final long CURRENT_TIME = System.currentTimeMillis();
```
basoko: Good catch!

rollbar-api/src/main/java/com/rollbar/api/payload/data/Person.java
```diff
@@ -15,10 +16,13 @@
 
   private final String email;
 
+  private final Map<String, Object> metadata;
+
   private Person(Builder builder) {
     this.id = builder.id;
     this.username = builder.username;
     this.email = builder.email;
+    this.metadata = builder.metadata != null ? unmodifiableMap(builder.metadata) : null;
```
mattdrees: Might not be a big deal, but `unmodifiableMap()` doesn't guarantee that the map is immutable. So this change actually changes Person from an immutable object to a mutable one (though it can only be mutated by mutating the map passed to the builder).

FWIW.

mattdrees:
@rokob is there anything I can do to help here? I am guessing you're wanting to solve the NPath complexity warnings that Codacy raised. If that's true, I can work on a solution.

PR #113, opened 2018-03-26
Add enabled flag to configuration and avoid processing if the notifier is disabled

### LINKED ISSUE #112: Allow Rollbar to be disabled ( e.g. in local development environments )
I would like to have this option: 
https://github.com/rollbar/rollbar-gem/blob/ffd4278332cc63b6dc29cde5fc3716793c455bec/lib/rollbar/configuration.rb#L19
available also in java plugin. 

Currently, there is no way to configure rollbar to be in "dryRun" mode. It is especially necessary in local development. Currently we need to specify some random rollbar access key like: "test" "local". It's not beautiful, especially when your servers replying with error that token is not valid. Also we generate load, which is completely dump.

We could however mock Rollbar here, but I personally don't want to have Mockito or JMock on my production classpath.

Regards
Mike.

basoko:
Fixes #112

### DIFF
--- a/rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
+++ b/rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
@@ -606,6 +606,11 @@ public class Rollbar {
     Config config = this.config;
     this.configReadLock.unlock();
 
+    if (!config.isEnabled()) {
+      LOGGER.debug("Notifier disabled.");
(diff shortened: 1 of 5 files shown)


rokob:
LGTM

Meganh1031:
[Megan Long.txt](https://github.com/rollbar/rollbar-java/files/1866682/Megan.Long.txt)

PR #114, opened 2018-04-09
Log4j2 integration

basoko:
This PR includes these changes:

* Make `Level.lookupByName` to be case insensitive.
* Add `ThrowableWrapper` to represent a `Throwable` that is not available.
* Deprecate in BodyFactory the `from` method that uses the `Throwable` in favor of the new `from` that uses the `ThrowableWrapper`.
* Overload in `Rollbar` the log method with a version that receives a `ThrowableWraper`.
* Add the log4j2 integration with the log4j2 appender.

### DIFF
--- a/rollbar-api/src/main/java/com/rollbar/api/payload/data/Level.java
+++ b/rollbar-api/src/main/java/com/rollbar/api/payload/data/Level.java
@@ -49,7 +49,7 @@ public enum Level implements Comparable<Level>, JsonSerializable {
   }
 
   public static Level lookupByName(String name) {
-    return name != null ? nameIndex.get(name) : null;
+    return name != null ? nameIndex.get(name.toLowerCase()) : null;
   }
 
   /**
--- a/rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
+++ b/rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
@@ -8,6 +8,8 @@ import com.rollbar.notifier.config.ConfigBuilder;
 import com.rollbar.notifier.config.ConfigProvider;
 import com.rollbar.notifier.uncaughtexception.RollbarUncaughtExceptionHandler;
 import com.rollbar.notifier.util.BodyFactory;
+import com.rollbar.notifier.wrapper.RollbarThrowableWrapper;
+import com.rollbar.notifier.wrapper.ThrowableWrapper;
 import java.lang.Thread.UncaughtExceptionHandler;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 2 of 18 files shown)


rollbar-api/src/main/java/com/rollbar/api/payload/data/Level.java
```diff
@@ -49,7 +49,7 @@
   }
 
   public static Level lookupByName(String name) {
-    return name != null ? nameIndex.get(name) : null;
+    return name != null ? nameIndex.get(name.toLowerCase()) : null;
```
rokob: 👍

rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
```diff
@@ -671,7 +697,8 @@ private Data buildData(Config config, Throwable error, Map<String, Object> custo
         .platform(config.platform())
         .language(config.language())
         .framework(config.framework())
-        .level(level != null ? level : level(error))
+        .level(level != null ? level : error != null ? level(error.getThrowable())
```
rokob: It is possible that `getThrowable` returns `null`, but that is okay here because `level(null)` gives the default.

rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
```diff
@@ -612,7 +637,8 @@ private void process(Throwable error, Map<String, Object> custom, String descrip
     }
     
     // Pre filter
-    if (config.filter() != null && config.filter().preProcess(level, error, custom, description)) {
+    if (config.filter() != null && config.filter().preProcess(level, error.getThrowable(), custom,
+        description)) {
       LOGGER.debug("Pre-filtered error: {}", error);
```
rokob: Here we used to debug log the throwable but now we will print the wrapper, not sure that matters.

rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
basoko: Yes, if it's a represented throwable, https://logging.apache.org/log4j/2.0/log4j-core/apidocs/org/apache/logging/log4j/core/impl/ThrowableProxy.html#getThrowable(). That was why i decided to introduce the `ThrowableWrapper` to represent this kind of errors where there is no access to the class of the Exception.

rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
basoko: Good point! After looking at it i think it's ok because we provide a `toString()` method of the `ThrowableWrapper` implementation (https://github.com/rollbar/rollbar-java/blob/log4j2-integration/rollbar-java/src/main/java/com/rollbar/notifier/wrapper/RollbarThrowableWrapper.java#L81) and potentially the `Throwable` source could not be available.

PR #116, opened 2018-04-10
Remove java.nio.charset.StandardCharsets dependency by replacing the UTF-8 charset by a string constant

### LINKED ISSUE #115: Crash on Android devices with API version lower than 19
The next line causes crash during rollbar initialization:
https://github.com/rollbar/rollbar-java/blob/f95c18e9406c3229adb4c22a29d51084b3ed8da8/rollbar-java/src/main/java/com/rollbar/notifier/sender/SyncSender.java#L25

This happens due to the class `StandardCharsets` does not exists in the Android 4.3 and lower.

basoko:
It seems that this is not available in Android 4.3 and lower.

Fix #115

### DIFF
--- a/rollbar-java/src/main/java/com/rollbar/notifier/sender/SyncSender.java
+++ b/rollbar-java/src/main/java/com/rollbar/notifier/sender/SyncSender.java
@@ -13,7 +13,6 @@ import java.io.OutputStream;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.nio.charset.StandardCharsets;
 
 /**
  * Synchronous implementation of the {@link Sender sender}.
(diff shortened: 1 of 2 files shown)


rollbar-java/src/main/java/com/rollbar/notifier/sender/SyncSender.java
```diff
@@ -22,7 +21,7 @@
 
   public static final String DEFAULT_API_ENDPOINT = "https://api.rollbar.com/api/1/item/";
 
-  private static final String UTF_8 = StandardCharsets.UTF_8.name();
```
rokob: The fact that the standard (correct) way to do this is `StandardCharsets.UTF_8.name()` is one the most Java things I have ever seen.

PR #119, opened 2018-04-13
Use buffered sender by default instead of the sync sender

### LINKED ISSUE #117: Make an async sender the default
- Use one of the queue based senders as the default sender instead of the current sync sender
- Document that the default is async and what the queueing strategy is
- Document how to replace the queue and the whole sender

basoko:
Fix #117

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,15 @@ compile('com.rollbar:rollbar-java:1.0.1')
 compile('com.rollbar:rollbar-android:1.0.1@aar')
 ```
 
+## How payloads are sent
+
+The actual notifier configuration builds a notifier that uses a BufferedSender to send the items
+to Rollbar. That sender is built using an unbound memory queue and a scheduled thread to send
+the events from the queue. 
+The queue as well as the frequency of the scheduled thread can be customized
+when building the buffered sender and it can be replaced by passing a custom configuration when 
+creating the notifier or initializing it.
+ 
 ## Usage
 
 For actual usage, the easiest way to get started is by looking at the examples:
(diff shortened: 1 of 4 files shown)


rokob:
So in other notifiers where we have an async send mechanism we also expose a function `wait` which when called stops the queue from accepting items and blocks until all of the items have been sent. I think that is beyond the scope of this PR but I just wanted to bring it up so I remember to add it later.

basoko:
No sure exactly the use case for that, but it seems that we could have the same behavior with a bounded Queue and some extra flags in the buffered sender to determine this behaviour. Now when the queue is full it just discard the element and notify about it through the `SenderListener`, so we could make it to block until there is more space.

rokob:
The use case is to delay shutdown until items have been processed; i.e. the use of sleep in your test would be replaced with a call to wait. But I forgot about the `SenderListener` so one could build that themselves.

basoko:
Ok! I see it now! Thanks for the explanation!

README.md
```diff
@@ -185,6 +185,15 @@ compile('com.rollbar:rollbar-java:1.0.1')
 compile('com.rollbar:rollbar-android:1.0.1@aar')
 ```
 
+## How payloads are sent
+
+The actual notifier configuration builds a notifier that uses a BufferedSender to send the items
+to Rollbar. That sender is built using an unbound memory queue and a scheduled thread to send
+the events from the queue. 
+The queue and as well the frequency of the scheduled thread can be customize
```
mbryzek: customize => customized

mbryzek: `The queue and` => the `and` should be removed

basoko: Thanks!

PR #125, opened 2018-04-21
Fix checkstyle errors and increase it from warning to error

basoko:
This PR increases the checkstyle errors from warning to error, that will cause the build to fail.


rokob:
Looks good

PR #143, opened 2018-05-14
Add support for configuring the rollbar url

zenlumi:
More than one rollbar url can be used by a company e.g. have different urls for different regions us, eu etc.

Added the **url** field in the `Config` interface to be further used in the _rollbar-logback_
module when setting a custom rollbar url (e.g. region specific rollar-us, rollbar-eu).

In _RollbarAppender_, added the possibility to customise the **environment**,
**language** (it can be `scala` for example) and to report the **server hostname**.

### DIFF
--- a/rollbar-java/src/main/java/com/rollbar/notifier/config/Config.java
+++ b/rollbar-java/src/main/java/com/rollbar/notifier/config/Config.java
@@ -27,6 +27,13 @@ public interface Config {
   String accessToken();
 
   /**
+   * Get the url.
+   *
+   * @return the Rollbar url.
+   */
+  String url();
+
+  /**
    * Get the environment.
    *
    * @return the environment.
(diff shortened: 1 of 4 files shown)


rollbar-java/src/main/java/com/rollbar/notifier/config/ConfigProviderHelper.java
```diff
@@ -0,0 +1,40 @@
...
+import java.lang.reflect.Constructor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Static helper for the ConfigProvider class.
+ */
+public class ConfigProviderHelper {
+
+  private static Logger thisLogger = LoggerFactory.getLogger("ConfigProviderHelper");
```
rokob: This should probably be `final`

PR #152, opened 2018-07-17
Do not throw “invalid access token” if it’s empty or is an spring-boot environment variable

celiovasconcelos:
How about avoiding this:

```java
com.rollbar.notifier.sender.exception.SenderException: com.rollbar.notifier.sender.exception.ApiException: invalid access token
	at com.rollbar.notifier.sender.AbstractSender.send(AbstractSender.java:41)
	at com.rollbar.notifier.sender.BufferedSender$SendTask.run(BufferedSender.java:217)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: com.rollbar.notifier.sender.exception.ApiException: invalid access token
	... 9 common frames omitted
```

When:

```xml
<appender name="ROLLBAR" class="com.rollbar.logback.RollbarAppender">
	<accessToken>${rollbar.api.key}</accessToken>
	...
</appender
```

Because:

Devs usually don't send exceptions to Rollbar when they are in development environment. Only the production server has the ${rollbar.api.key} variable. This code helps to avoid using that boring `janino ` ifs...

### DIFF
--- a/rollbar-logback/src/main/java/com/rollbar/logback/RollbarAppender.java
+++ b/rollbar-logback/src/main/java/com/rollbar/logback/RollbarAppender.java
@@ -62,24 +62,29 @@ public class RollbarAppender extends AppenderBase<ILoggingEvent> {
   @Override
   public void start() {
 
-    ConfigProvider configProvider = ConfigProviderHelper
-            .getConfigProvider(this.configProviderClassName);
-    Config config;
-
-    ConfigBuilder configBuilder = withAccessToken(this.accessToken)
-            .environment(this.environment)
-            .endpoint(this.endpoint)
-            .server(new ServerProvider())
-            .language(this.language);
-
-    if (configProvider != null) {
-      config = configProvider.provide(configBuilder);
-    } else {
-      config = configBuilder.build();
+    if (this.accessToken != null && !this.accessToken.endsWith("IS_UNDEFINED")){
+      
+      ConfigProvider configProvider = ConfigProviderHelper
+              .getConfigProvider(this.configProviderClassName);
+      Config config;
+  
+      ConfigBuilder configBuilder = withAccessToken(this.accessToken)
+              .environment(this.environment)
+              .endpoint(this.endpoint)
+              .server(new ServerProvider())
+              .language(this.language);
+  
+      if (configProvider != null) {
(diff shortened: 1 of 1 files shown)


basoko:
Hi @celiovasconcelos,

I think that instead of modifying the notifier another alternative to solve the problem you mention would be to create in your end an implementation of a [`ConfigProvider`](https://github.com/rollbar/rollbar-java/blob/master/rollbar-java/src/main/java/com/rollbar/notifier/config/ConfigProvider.java) that enables or disables the notifier based on the access token value or other logic you want.

Then could you use your `ConfigProvider` by setting the `configProviderClassName`, in the logback configuration file, to your actual class name.

```
<appender name="ROLLBAR" class="com.rollbar.logback.RollbarAppender">
	<accessToken>${rollbar.api.key}</accessToken>
	<configProviderClassName>(your_config_provider_class_name)</configProviderClassName>
</appender
```

Hope this can help you, and thanks for contributing!

celiovasconcelos:
Thank you for your answer.
I would prefer "convention over configuration". 🤓

basoko:
Glad to hear that it worked!

I agree about the "convention over configuration", but defaulting to not do anything maybe seems ok for a dev environment meanwhile in a production one, not having the access token, maybe due to a mistake in the configuration files, it's an indicator of an error somewhere and knowing about it as fast as possible, seems better.

Other way to not have the notifier working in a dev environment would be about using different `logback` configuration files per environment, as I guess you would also like to have different logging levels depending on the environment, so for the dev one, you could have that conf file without this appender.

Thanks!

Issue #167, opened 2018-10-26
(rollbar-android) Messages sent twice when custom sender is used


mannodermaus:
Hi, thanks for the SDK. I'm encountering an issue with applying a custom sender to my `Rollbar` instance, which results in all messages being reported twice.

1) `com.rollbar.android.Rollbar` initializes a default `BufferedSender` [in its constructor](https://github.com/rollbar/rollbar-java/blob/master/rollbar-android/src/main/java/com/rollbar/android/Rollbar.java#L249:L253)
2) A `BufferedSender` immediately schedules its queue observer [in its constructor](https://github.com/rollbar/rollbar-java/blob/master/rollbar-java/src/main/java/com/rollbar/notifier/sender/BufferedSender.java#L54:L55)
3) When configuring my custom configuration for Rollbar, I provide another sender that observes the same cache directory:
```java
Rollbar.init(context,
        accessToken,
        environment,
        true,
        true,
        (builder) -> builder.sender(new CustomSender()).build()
);
```
4) Both the default `BufferedSender` and my `CustomSender` now poll from the same folder, which results in messages potentially being reported twice (note the groupings of 2 in the screenshot below):
<img width="217" alt="screen shot 2018-10-26 at 8 42 29" src="https://user-images.githubusercontent.com/3897705/47549036-1ecde100-d8fb-11e8-8d2a-fc9a40b45e2c.png">

After closing the app, the default sender isn't closed, and its thread remains active as well.

A fix could be to explicitly tell the `Sender` to start observing *after* all configuration has been applied.

**Expected:**
When a custom `sender()` is used, the default instance should stop observing

**Actual:**
The default `Sender` doesn't stop observing when a custom one is applied

**Version:**
`rollbar-android:1.3.1`

PR #168, opened 2018-10-26
Close default Sender if custom implementation is used

mannodermaus:
The default `Sender` will start scheduling its queue observer right in the constructor, which may conflict with a custom `Sender` provided through `ConfigProvider`. If the configured instance doesn’t match the default, the latter is closed again immediately.

Resolves https://github.com/rollbar/rollbar-java/issues/167 by closing the default instance when a custom one is given, too.

### DIFF
--- a/rollbar-android/src/main/java/com/rollbar/android/Rollbar.java
+++ b/rollbar-android/src/main/java/com/rollbar/android/Rollbar.java
@@ -17,7 +17,6 @@ import com.rollbar.notifier.config.Config;
 import com.rollbar.notifier.config.ConfigBuilder;
 import com.rollbar.notifier.sender.BufferedSender;
 import com.rollbar.notifier.sender.queue.DiskQueue;
-import com.rollbar.notifier.util.ObjectsUtils;
 
 
 import java.io.File;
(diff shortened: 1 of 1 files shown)


basoko:
Thanks @mannodermaus for these changes!

I think that instead of closing the default sender, if a custom one is provided, it'd be better to not create it until it's verified that there isn't a custom one.

mannodermaus:
I initially decided against that approach because of the awkward re-creation of a `ConfigBuilder`, but I agree that this would probably be better.

This is what this could look like:

```java
    ConfigBuilder defaultConfig = ConfigBuilder.withAccessToken(accessToken)
        .client(clientProvider)
        .platform(ANDROID)
        .framework(ANDROID)
        .notifier(new NotifierProvider(NOTIFIER_VERSION))
        .environment(environment)
        .handleUncaughtErrors(registerExceptionHandler);

    Config config;
    if (configProvider != null) {
      config = configProvider.provide(defaultConfig);
    } else {
      config = defaultConfig.build();
    }

    if (config.sender() == null) {
      DiskQueue queue = new DiskQueue.Builder()
              .queueFolder(folder)
              .build();

      BufferedSender sender = new BufferedSender.Builder()
              .queue(queue)
              .initialFlushDelay(TimeUnit.SECONDS.toMillis(DEFAULT_ITEM_SCHEDULE_STARTUP_DELAY))
              .flushFreq(TimeUnit.SECONDS.toMillis(DEFAULT_ITEM_SCHEDULE_DELAY))
              .build();

      config = ConfigBuilder.withConfig(config)
              .sender(sender)
              .build();
    }
```

basoko:
I think this approach would  be better, as we avoid creating and closing a sender that is not going to be used, although also agree about that awkward re-creation of the`ConfigBuilder`.

Thanks @mannodermaus!

basoko:
Thanks @mannodermaus!

@rokob I think this could be merged and released if you are ok with the changes.

rokob:
LGTM

PR #169, opened 2018-12-03
Get local variables via the JVMTI

### LINKED ISSUE #130: Capture local variables
Similar to pyrollbar. See 'locals' here: https://docs.rollbar.com/docs/python#section-configuration-reference

rokob:
Still needs some testing and some more work, but I wanted to get this up here as a PR sooner rather than later.

Fixes #130


native-agent/src/env.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    pub fn get_local_object(
+        &mut self,
+        thread: jthread,
+        depth: jint,
+        slot: jint,
+        result: *mut jobject,
+    ) -> Result<()> {
+        jvmtifn!(self.jvmti, GetLocalObject, thread, depth, slot, result)
```
rokob: This is the only part that I'm not 100% on. The docs say the object returned in the result pointer here is a JNI local reference which must be "managed". I am pretty sure we are in the clear here, but if there were to be a problem it would come from this.

PR #179, opened 2019-03-07
Add configuration to determine default item level

### LINKED ISSUE #176: Add config option to override default Throwable log level
It's currently defaulting to Critical

rokob:
Fixes #176

By default we use WARNING for messages, CRITICAL for Error type
throwables, and ERROR for other throwables. There is no way to change
that than always setting the level. This makes it so you can set the
default level in the configuration.

### DIFF
--- a/rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
+++ b/rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
@@ -618,9 +618,13 @@ public class Rollbar {
   }
 
   /**
-   * Get the level of the error or message. By default: CRITICAL for {@link Error}, ERROR for other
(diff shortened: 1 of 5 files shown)


rollbar-java/src/main/java/com/rollbar/notifier/Rollbar.java
```diff
@@ -624,14 +624,14 @@ public void log(ThrowableWrapper error, Map<String, Object> custom, String descr
    * @param error the error.
    * @return the level.
```
rokob: Fix these docs

rokob: ✅

PR #180, opened 2019-04-10
Enable sending complete JSON payloads, as proxy for other SDKs

waltjones:
Enables the public interface for rollbar-android: `void sendJsonPayload(String json)`

This supports Cordova apps by allowing the Java env to handle sending Rollbar events for the JS env. This also means each of the SDKs (At least Java, iOS, and rollbar.js) should agree on the interface definition, and be able to support it.

### DIFF
--- a/rollbar-android/src/main/java/com/rollbar/android/Rollbar.java
+++ b/rollbar-android/src/main/java/com/rollbar/android/Rollbar.java
@@ -794,6 +794,15 @@ public class Rollbar {
   }
 
   /**
+  * Send payload to Rollbar.
+  *
+  * @param payload JSON payload string.
+  */
+  public void sendJsonPayload(String json) {
+    rollbar.sendJsonPayload(json);
+  }
+
+  /**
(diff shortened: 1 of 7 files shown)


rokob:
Do we want to do all the work somewhere else and have this just be a proxy or do we want to send structured data over here and let this SDK handle truncation, scrubbing, etc? For what this is going to be used for just the proxy route seems fine for now, but we might want to think how this could be done to have the native code do more of the heavy lifting.

ArturMoczulski:
LGTM as well :)

waltjones:
@rokob I'd love to discuss pros/cons of where would be optimal. After conversion to a string is the lowest marshaling complexity, though it would still be possible to do truncation after that point. I'm not sure I'm seeing a clear benefit. 

One benefit of getting as close to the API request as possible is that any of the SDKs could probably conform to this pattern. It was certainly a win that at least JS, Java, and iOS all could.

Are there specific parts that could be handled better outside the environment where the logging and error handling happen?

PR #185, opened 2019-05-02
Expose `RollbarAppender(rollbar Rollbar)` constructor overload, and make `RollbarAppender#start()` skip `this.rollbar` initialization if it's already non-null

### LINKED ISSUE #184: Make `RollbarAppender(Rollbar rollbar)` constructor public and allow `start()` to skip `Rollbar` instantiation so `RollbarAppender` can be constructed and started programmatically
Similar to #182, I'd like to construct and start a `RollbarAppender` programmatically. As I see it, there are two things preventing this:
* I already have a `Rollbar` instance that I'd like to inject, but the constructor is package-private; making it public as in #182 would fix this.
* I'd need to call `RollbarAppender.start()` myself, so make that function skip `this.rollbar` instantiation if it already exists.

simonyun:
Fixes #184

### DIFF
--- a/rollbar-logback/src/main/java/com/rollbar/logback/RollbarAppender.java
+++ b/rollbar-logback/src/main/java/com/rollbar/logback/RollbarAppender.java
@@ -47,11 +47,11 @@ public class RollbarAppender extends AppenderBase<ILoggingEvent> {
   private String configProviderClassName;
 
   /**
-   * Constructor for testing purposes.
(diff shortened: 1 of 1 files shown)


ArturMoczulski:
Released now in https://github.com/rollbar/rollbar-java/releases/tag/v1.5.2

simonyun:
thank you @ArturMoczulski!! much appreciated.

PR #190, opened 2019-08-15
Fix pre filter error NullPointerException

nilsjr:
I tweaked out that some functionality like log info and warning doesn't work with the current version.

since 9c3e31db484cad5336ee984a877a62150bb1761c info, warnings and other logs without an error (null argument) will result in an internal exception in following line:

```
 private void process(ThrowableWrapper error, Map<String, Object> custom, String description,
      Level level, boolean isUncaught) {
    ...
    // Pre filter
    if (config.filter() != null && config.filter().preProcess(level, error.getThrowable(), custom,
        description)) {
      LOGGER.debug("Pre-filtered error: {}", error);
      return;
    }
    ...
}
```

Due error could be null the wrapper getThrowable() call will throw a null pointer exception.

**actual:**
rollbar.info("test") -> doesn't appears in rollbar dashboard

**expected:**
rollbar.info("test") -> does appear in rollbar dashboard

### Additional:
I also added nullable annotation to filter preProcess function due better kotlin usage. 

actual:
```
preProcess(
            level: Level,
            error: Throwable,
            custom: MutableMap<String, Any>,
            description: String
          )
```
This results in an illegal argument exception if java calls the interface with a null argument. 
E.g. preProcess(.., null, null, null) -> Exception

now with nullable annotation we have to implement like this:
```
object : Filter {
          override fun postProcess(data: Data?): Boolean {
            return false
          }

          override fun preProcess(
            level: Level,
            error: Throwable?,
            custom: MutableMap<String, Any>?,
            description: String?
          ): Boolean {
            return (error is UnknownHostException)
          }
}
```

### DIFF
--- a/rollbar-java/build.gradle
+++ b/rollbar-java/build.gradle
@@ -16,6 +16,8 @@ dependencies {
 
     compile group: 'org.slf4j', name: 'slf4j-api', version: '1.7.25'
 
+    compile 'com.google.code.findbugs:jsr305:3.0.2'
+
     integTestCompile group: 'com.github.tomakehurst', name: 'wiremock', version: '2.11.0'
     integTestCompile group: 'com.google.code.gson', name: 'gson', version: '2.8.2'
 }
--- a/rollbar-java/build.gradle
+++ b/rollbar-java/build.gradle
@@ -16,7 +16,7 @@ dependencies {
 
     compile group: 'org.slf4j', name: 'slf4j-api', version: '1.7.25'
 
-    compile 'com.google.code.findbugs:jsr305:3.0.2'
+    compileOnly 'com.google.code.findbugs:jsr305:3.0.2'
 
     integTestCompile group: 'com.github.tomakehurst', name: 'wiremock', version: '2.11.0'
     integTestCompile group: 'com.google.code.gson', name: 'gson', version: '2.8.2'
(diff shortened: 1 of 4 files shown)


rollbar-java/build.gradle
```diff
@@ -16,6 +16,8 @@ dependencies {
 
     compile group: 'org.slf4j', name: 'slf4j-api', version: '1.7.25'
 
+    compile 'com.google.code.findbugs:jsr305:3.0.2'
```
basoko: Why is this dependency added?

nilsjr:
Add test that verifies filter.preProcess is called if exception is null

rollbar-java/build.gradle
basoko: I think we change that dependency to `compileOnly`? You will need to include it from your project if Kotlin needs it at runtime.

basoko:
Thanks for adding the test @nilsjr !

The only thing left is the question about changing the scope of that dependency you added.I think we could change that dependency to `compileOnly`,  and any project that needs the `javax.annotation.Nullable` at runtime could provide it.

nilsjr:
Yeah that feedback I already heard from my colleague. Thanks for your feedback

basoko:
Thanks @nilsjr for this fix!! I think @ArturMoczulski we should merge it and make a release of it.

nilsjr:
@ArturMoczulski @basoko Five months later, still no merge 😢

ArturMoczulski:
Released in https://github.com/rollbar/rollbar-java/releases/tag/v1.6.0

PR #198, opened 2020-02-11
Fix for a 403 error response from Nginx

ZenAnnard:
It seems possible to access Rollbar's servers and get a 403 Forbidden response.
The code didn't handle this well at all since it always expects a json response
from the server.
This patch checks if the regex matcher succeeded and gracefully returns an error
response so you can actually determine what is going on.


Macworld34:
Great that for that problem $olve

PR #212, opened 2020-05-11
Adding more clarity into the native-agent README

vselvarajijay:


native-agent/README.md
```diff
@@ -23,7 +23,21 @@ Gradle the easiest way is to add the following to your `build.gradle` file:
 applicationDefaultJvmArgs = ["-agentpath:path/to/"+System.mapLibraryName("rollbar_java_agent")]
 ```
 
-Regardless of your JVM language of choice, at some level their is an invocation of the JVM and
+Once you have your Java code using the Native Agent, you need make sure your `rollbar-java` is
+configured with appPackages at a minimum. If you also want to handle uncaught exceptions,
+make sure the handleUncaughtErrors is set to true.
+
+```
+Rollbar.init(withAccessToken("ACCESS-TOKEN")
```
basoko, line 31: I'd mention that the kind of values expected for `appPackages` are packages prefix like `com.example.app` and that is required to track information about exceptions that have class names matching those prefixes in the stack trace.

vselvarajijay: @basoko - let me know if the new update looks better

basoko: In that code snippet I'd hange to use the `new` operator instead of using the `Rollbar.init` and just add a specific example of appPackages like:
```
new Rollbar(withAccessToken("ACCESS-TOKEN")
     .appPackages(Arrays.asList("com.example.app"))
    .handleUncaughtErrors(true)
    .build());

PR #214, opened 2020-05-17
Remove VERSION constant in rollbar-android and load from metadata

basoko:
A new metadata property has been added with the name
`com.rollbar.android._notifier.version` in the library AndroidManifest
that will be injected from gradle using manifest place holders.

This will help to handle the versioning as now the value used 
is the `VERSION_NAME` coming from `gradle.properties`.

### DIFF
--- a/rollbar-android/build.gradle
+++ b/rollbar-android/build.gradle
@@ -30,6 +30,7 @@ android {
         minSdkVersion 16
         targetSdkVersion 27
         consumerProguardFiles 'proguard-rules.pro'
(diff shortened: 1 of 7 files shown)


basoko:
@waltjones my idea was denote it, somehow, that the `notifier` part of the namespace `com.rollbar.android` is private or for internal use. As opposed to `com.rollbar.android.ACCESS_TOKEN` that is supposed to be provided by clients.

waltjones:
Got it. Makes sense. :+1:

PR #216, opened 2020-05-20
Rollbar Spring Boot SDK

vselvarajijay:
Native integration of Rollbar for Spring Boot


examples/rollbar-spring-boot-webmvc/src/main/java/com/example/springbootwebmvc/RollbarConfig.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.web.servlet.config.annotation.EnableWebMvc;
+
+
+@Configuration()
+@EnableWebMvc
+@ComponentScan({"com.example.springbootwebmvc","com.rollbar.spring.boot.webmvc"})
```
basoko, line 15: I think that we can just use here this:

```
@ComponentScan({"com.example.springbootwebmvc","com.rollbar.spring"})
```

If not, in order to make Spring discover the `RollbarHandlerExceptionResolver` we should add to that list the `com.rollbar.spring.webmvc` that would be more verbose.

examples/rollbar-spring-boot-webmvc/src/main/resources/application.properties
```diff
@@ -0,0 +1,3 @@
+rollbar.access_token = <ACCESS_TOKEN>
+rollbar.environment = development
+rollbar.framework = spring-webmvc
```
basoko, line 3: I think this param is not used, and it will be filled up by our default config in Spring.

rollbar-spring-boot-webmvc/src/main/java/com/rollbar/spring/boot/webmvc/RollbarSpringBootExceptionResolver.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.web.servlet.ServletListenerRegistrationBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+
+@Order(Ordered.HIGHEST_PRECEDENCE)
+@ControllerAdvice
+public class RollbarSpringBootExceptionResolver extends RollbarHandlerExceptionResolver {
```
basoko, line 15: We don't need to implement a `HandlerExceptionResolver` we can just use the one that we have packaged in `rollbar-spring-webmvc`. 
What we need to do for Spring boot is to make it aware about a `ServletContextInitializer` implementation, so we can register the `RollbarRequestListener` there, because seems that with Spring boot the `WebApplicationInitializer` is not called.

examples/rollbar-spring-boot-webmvc/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    compile project(':rollbar-spring-boot-webmvc')
+    implementation 'org.springframework.boot:spring-boot-starter-web:' + springBootStarterVersion
+
+    testImplementation('org.springframework.boot:spring-boot-starter-test') {
```
basoko, line 19: I think we don't need this dependency.

examples/rollbar-spring-boot-webmvc/src/main/java/com/example/springbootwebmvc/RollbarConfig.java
vselvarajijay: Looks like we need the springbootwebmvc so that the ServletContextInitializer gets called in the spring-boot app

```Java
@ComponentScan({
        "com.example.springbootwebmvc",
        "com.rollbar.spring.boot.webmvc",
        "com.rollbar.spring.webmvc"
})
```

rollbar-spring-boot-webmvc/src/main/java/com/rollbar/spring/boot/webmvc/RollbarSpringBootExceptionResolver.java
vselvarajijay: Ok i refactored this in the next commit

vselvarajijay: Btw, as part of this I removed the unit-test for RollbarSpringBootExceptionResolver. I don't think the RollbarSpringBootConfiguration class I added has much to test. It just registers a bean, and adds the RollbarRequestListener.class to the servlet listener.

rollbar-spring-boot-webmvc/src/main/java/com/rollbar/spring/boot/webmvc/RollbarSpringBootConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+import com.rollbar.web.listener.RollbarRequestListener;
+
+import javax.servlet.ServletContext;
+
+import org.springframework.boot.web.servlet.ServletContextInitializer;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class RollbarSpringBootConfiguration {
```
basoko, line 12: I don't understand why we want to provide this `Configuration` in this way, that registers a bean by creating an anonymous class of a `ServletContextInitializer`.

I think we can just provide an implementation of a `ServletContextInitializer` that is the piece needed. Also, I think that if you use the `@Component` annotation it will get registered by using the `@ComponentScan({"com.rollbar.spring"})`

examples/rollbar-spring-boot-webmvc/src/main/java/com/example/springbootwebmvc/RollbarConfig.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.web.servlet.config.annotation.EnableWebMvc;
+
+
+@Configuration()
+@EnableWebMvc
+@ComponentScan({
+        "com.example.springbootwebmvc",
+        "com.rollbar.spring.boot.webmvc",
```
basoko, line 17: IMHO I think it's just clearer to specify as a package to scan `com.rollbar.spring` and the client's package of the integration, that in this case is this example project. I'd opt for this:

```
@ComponentScan({
        "com.example.springbootwebmvc",
        "com.rollbar.spring"
})
```

As this level of detail does not provide any value and may cause some confusion.

examples/rollbar-spring-boot-webmvc/src/main/java/com/example/springbootwebmvc/RollbarConfig.java
vselvarajijay: Disregard I was able to use the simpler method:

```java
@ComponentScan({"com.example.springbootwebmvc","com.rollbar.spring"})
```

vselvarajijay:
+ Native integration of Rollbar for Spring Boot

PR #249, opened 2021-02-25
Move all Travis build steps to Github action.

diegov:
## Description of the change

This change adds an additional `ci.yml` GH action, which reproduces as closely as possible all the steps that were being done in Travis CI, and removes the Travis CI configuration.

## Type of change
- [x] New **feature** (non-breaking change that adds functionality)

## Related issues

None
 
### Development

- [x] Lint rules pass locally

Note there are no automated tests for the GH action nor the Travis configuration, but it has been extensively tested manually, locally where possible, and via numerous pushes to the repo.

### Code review 


waltjones (approved):
:+1:

.github/workflows/ci.yml
```diff
@@ -0,0 +1,117 @@
...
+          GPG_KEY_ID: ${{ secrets.GPG_KEY_ID }}
+          GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
+          ENCRYPTED_GPG_KEY_LOCATION: ./.github/secring.gpg.enc
+          GPG_KEY_LOCATION: ./.github/secring.gpg
+          NEXUS_USERNAME: ${{ secrets.NEXUS_USERNAME }}
+          NEXUS_PASSWORD: ${{ secrets.NEXUS_PASSWORD }}
+        run: |
+          openssl enc -aes-256-cbc -K "$SECRING_GPG_KEY" -iv "$SECRING_GPG_IV" -in "$ENCRYPTED_GPG_KEY_LOCATION" -out "$GPG_KEY_LOCATION" -d &&
+          source ~/.jdk7_home &&
+          ./.github/release.sh
```
basoko, line 112: What do you think about having a specific build step in the release job, instead of executing it as a dependency of the publish task?

basoko (approved):
Nice work @diegov ! I left a couple of minor comments!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,117 @@
...
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ matrix.java }}-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-${{ matrix.java }}-
+
+      - name: Environment info
+        run: |
+          ./gradlew --version
+          echo ANDROID_SDK_ROOT: ${ANDROID_SDK_ROOT}
```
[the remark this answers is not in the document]
diegov: The following SDK versions and build tools are installed in the base image:

https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#android

This includes compile and target SDK 27, and build tools 28.0.3, which is what we set in our `build.gradle`. 

It might still be a good idea to explicitly install them even if they are currently available in the base image, since GH has already removed SDK versions from the base image, and it could eventually remove a version we need. I'll see if we can do the install on the same dir the base image puts the SDKs, that way if the versions we need are already there, it won't add much time to the build.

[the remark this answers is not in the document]
diegov: It turns out gradle will install any missing SDK, so that's our fallback if the base image drops one of the versions we need. No need to update the PR

.github/workflows/ci.yml
diegov: The build task runs in the `build` matrix job (currently using only java 8, but the next PR will add java 11). That's a different job and build environment from the release, which is different from Travis where build and release where running in the same build environment. But in the Travis case the release happened after `clean` and `check`, so it was also relying on the task dependencies to get everything built for publishing.

basoko: Yes, sorry, what i meant is if it would be worth having a specific step in the release to make the specific build task that is going to be released. I was thinking that maybe if you have to take a look at the output, having that explicit step could be better than looking at the output of the `release.sh` script.

PR #254, opened 2021-03-09
Catch Throwable in payload send task, to ensure it doesn't stop being…

### LINKED ISSUE #246: Rollbar stops sending after few days
We upgraded our java server recently to the recent spring libraries, and rollbar stops sending errors after few days of the server working. Nothing is found in our logs. Is there any way to log locally when rollbar fails to send?
We have other rollbar instances, not in java, that are sending as usual. We also have other java servers that are sending as usual but they are not in production and they are restarted very frequently.

diegov:
## Description of the change

This change represents a compromise solution to handling errors that are categorised as fatal, but that often are not. The issue that prompted this, #246, was a result of us *not* catching a VirtualMachineError instance (stack overflow in that case), as per the Java spec recommendation, but then the JVM continuing to run without any visible problem, which means from our user's perspective their application was running fine except Rollbar stopped sending occurrences. 

After looking into OpenJDK's implementation of FutureTask and other asynchronous classes, there are several scenarios where it internally catches Throwable, without even re-throwing. That was actually happening in our issue, since we were not catching the stack overflow but it was being caught two stack frames above our method, with no special handling of any sort. 

Given that Throwable was being caught already via standard JDK usage, it is no less safe if we do it ourselves, and by doing it we can ensure our task keeps running.

On top of basic unit testing, this has been manually tested under HotSpot 8 and 11 and J9 11 JVMs, 

## Type of change
- [x] Bug fix (non-breaking change that fixes an issue)

## Related issues

> Fix #246 

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/rollbar-java/src/main/java/com/rollbar/notifier/sender/BufferedSender.java
+++ b/rollbar-java/src/main/java/com/rollbar/notifier/sender/BufferedSender.java
@@ -250,6 +250,19 @@ public class BufferedSender implements Sender {
         for (SenderListener senderListener : sender.getListeners()) {
           senderListener.onError(payload, new SenderException(e));
         }
+      } catch (Throwable e) {
+        // Catching all throwables is not great, but if we don't catch it here, FutureTask catches
+        // it a couple of stack frames above us, sets itself as failed, and the executor service
+        // won't schedule our task again. FutureTask does not re-throw either, so there is no
+        // difference in that regard with us catching the error.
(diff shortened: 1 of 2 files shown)


basoko (approved):
Very nice @diegov ! :100:

Az111z2:
A

Az111z2 (approved):

PR #255, opened 2021-03-09
Revert "Catch Throwable in payload send task, to ensure it doesn't stop being…"

Az111z2:
Reverts rollbar/rollbar-java#254


diegov:
Hi @Az111z2, could you add some context for why you think we should revert #254?

PR #261, opened 2021-03-24
Don't require an access token in log4j appender when configuration class is provided

diegov:
## Description of the change

The configuration class can provide the access token, so we don't need it to be required as an appender parameter when a configuration class is configured. 

This change will validate that either a configuration class or an access token is provided (both are also allowed), and will fail to initialise the appender *only* when neither is present. 

## Type of change
- [x] Bug fix (non-breaking change that fixes an issue)

## Related issues

None

### Development

- [x] Lint rules pass locally

### Code review 


waltjones (approved):
:+1:

PR #262, opened 2021-04-09
Add revapi binary and source backwards compatibility check to build.

diegov:
## Description of the change

This adds a revapi (https://github.com/revapi/revapi) binary and compile level compatibility check to the gradle build.

Revapi seems to be the most straight forward and maintained Java binary compatibility checker available. Based on my tests it has very good (if not complete) coverage of the Java spec's "Binary Compatiblity" section (https://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html)

The gradle plugin will automatically check the build against the most recent git tag, which works very well in our case since all published versions have tags. It will skip checks when a published artifact does not exist for a given project, so it won't break when new projects are added, but will automatically start checking them as soon as we have published an initial version.

Breaking changes can be added to an exclusion list, so it won't prevent us from making changes intentionally, but hopefully it will catch accidental compatibility breakages. 

## Type of change

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/.github/fetch_to_tag.sh
+++ b/.github/fetch_to_tag.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+set -e
+
+# Github actions will do shallow clones which is great for performance, but some of our
+# tasks require a history going back to the most recent tag, so this script achieves
+# that, without pulling the full history
+
+function fetch_to_tag() {
+    n="$1"
+    if [ "$n" == "" ]; then
+        n=2
+    fi
+    
+    echo "Fetching commits with depth ${n}"
+
+    rev=$(git rev-list -n 1 HEAD)
+
+    git fetch origin --depth "$n" "$rev"
+
+    if ! git describe --tags --abbrev=0 HEAD^ 1>/dev/null 2>&1; then
+        fetch_to_tag "$(expr "$n" \* 2)"
+    fi
+}
+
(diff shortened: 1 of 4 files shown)


diegov:
Related to [ch83040]

basoko (commented):
Nice @diegov ! One question, will this task be executed in every PR? What would happen if a PR breaks the compatibility? Is there any plan for handling those cases?

diegov:
> Nice @diegov ! One question, will this task be executed in every PR? What would happen if a PR breaks the compatibility? Is there any plan for handling those cases?

It will run for each PR. If a PR breaks compatibility, and we're OK with it, there is a gradle command which is part of the plugin, to accept all breaking changes:

```
./gradlew :project:revapiAcceptAllBreaks
```

Or alternatively, each change can be accepted individually:

```
./gradlew revapiAcceptBreak --justification "{why this is ok}" \
        --code "{revapi check code}" \
        --old "{optional revapi description of old element}" \
        --new "{optional revapi description of new element}"
```

These commands create or update the project's ignore file, which must be added to the commit, and makes the check pass. 

I think this is a pretty decent workflow overall.

diegov:
The commands I pasted above are written to the build log (with the specific details of the error, not as templates) as part of the error if the build fails, so it's clear what to do when it fails.

basoko (approved):
Thanks for the explanation! Very cool!

PR #264, opened 2021-05-20
Use generated class instead of jar manifest to populate notifier vers…

diegov:
This PR updates the method we use to find the notifier version, to use a generated Java class instead of the jar's manifest attributes. 

This fixes the notifier version attribute when users shade rollbar-java, since until now we could end up reporting the version of the user's application as the notifier version instead of our own, due to the fact that our classes had been merged into the user's jar, and only one manifest resource (the user's) remained. 

The approach we take now is to generate a static class at build time, with a single method returning the version. This removes the need for any dynamic resource loading, which could fail under relocation since the references are strings. 

## Description of the change

> Description here
## Type of change
- [X] Bug fix (non-breaking change that fixes an issue)

## Related issues

[ch85159]

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,6 @@ atlassian-ide-plugin.xml
 
 # Secrets
 .github/secring.gpg
+
+# Local gradle properties
+local.properties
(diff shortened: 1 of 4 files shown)


diegov:
Wrong clubhouse reference 🤦‍♂️

PR #265, opened 2021-05-20
Use generated class instead of jar manifest to populate notifier version

diegov:
## Description of the change

This PR updates the method we use to find the notifier version, to use a generated Java class instead of the jar's manifest attributes. 

This fixes the notifier version attribute when users shade rollbar-java, since until now we could end up reporting the version of the user's application as the notifier version instead of our own, due to the fact that our classes had been merged into the user's jar, and only one manifest resource (the user's) remained. 

The approach we take now is to generate a static class at build time, with a single method returning the version. This removes the need for any dynamic resource loading, which could fail under relocation since the references are strings. 

Automated test coverage isn't great on this change, but I ran a shading + relocation test using the `com.github.johnrengelman.shadow` gradle plugin, with relocated classes:

```groovy
shadowJar {
    relocate 'com.rollbar', 'c.rb'
    archiveBaseName.set('shadow')
    archiveClassifier.set('')
    archiveVersion.set('')


    manifest {
        attributes(
                "Implementation-Title": project.name,
                "Implementation-Version": VERSION_NAME
        )
    }
}

```

I've verified that the classes are actually relocated in the Jar:
```
 $ jar tf shadow.jar  | grep c/rb/notifier/provider/notifier/VersionHelper
c/rb/notifier/provider/notifier/VersionHelper.class
c/rb/notifier/provider/notifier/VersionHelperResources.class
```

Using `rollbar-java` 1.7.6:

```json
{
    "access_token": "token",
    "data": {
        "body": {
            "message": {
                "body": "test"
            }
        },
        "code_version": "0.9.55_APP",
        "language": "java",
        "level": "error",
        "notifier": {
            "name": "rollbar-java",
            "version": "0.9.55_APP"
        },
        "timestamp": 1621971709.178
    }
}
```

With this fix:

```json
{
    "access_token": "token",
    "data": {
        "body": {
            "message": {
                "body": "test"
            }
        },
        "code_version": "0.9.55_APP",
        "language": "java",
        "level": "error",
        "notifier": {
            "name": "rollbar-java",
            "version": "1.7.7-SNAPSHOT"
        },
        "timestamp": 1621971744.911
    }
}
```

## Type of change

- [X] Bug fix (non-breaking change that fixes an issue)

## Related issues

[ch85159]

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/rollbar-java/build.gradle
+++ b/rollbar-java/build.gradle
@@ -38,9 +38,8 @@ task createVersionClass() {
     def pkg = ["com", "rollbar", "notifier", "provider", "notifier"];
 
     def pkgName = pkg.join(".");
-    def pkgPath = "${ext.OUTPUT_DIR}${File.separator}" +
-            (['src', 'generated'] + pkg).join(File.separator)
-    def escapedVersion = VERSION_NAME.replace('"', '\\"');
+    def pkgPath = "${ext.OUTPUT_DIR}${File.separator}${pkg.join(File.separator)}"
+    def escapedVersion = VERSION_NAME.replace('\\', '\\\\').replace('"', '\\"');
 
     def classText = """package ${pkgName};
 
--- a/rollbar-java/build.gradle
+++ b/rollbar-java/build.gradle
@@ -12,12 +12,66 @@ buildscript {
 apply plugin: "nebula.integtest"
 
 dependencies {
-    api project(':rollbar-api')
+  api project(':rollbar-api')
 
-    api group: 'org.slf4j', name: 'slf4j-api', version: '1.7.25'
+  api group: 'org.slf4j', name: 'slf4j-api', version: '1.7.25'
 
-    compileOnly 'com.google.code.findbugs:jsr305:3.0.2'
+  compileOnly 'com.google.code.findbugs:jsr305:3.0.2'
 
-    integTestImplementation group: 'com.github.tomakehurst', name: 'wiremock', version: '2.27.0'
-    integTestImplementation group: 'com.google.code.gson', name: 'gson', version: '2.8.2'
+  integTestImplementation group: 'com.github.tomakehurst', name: 'wiremock', version: '2.27.0'
+  integTestImplementation group: 'com.google.code.gson', name: 'gson', version: '2.8.2'
+}
+
+/**
+ * This task will create a version property that is statically referenced when populating the
+ * `notifier` section of the payload. It helps when users shade and / or relocate the
+ * `rollbar-java` classes, since in those cases we no longer have access to our jar manifest.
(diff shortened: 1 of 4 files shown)


rollbar-java/build.gradle
```diff
@@ -12,12 +12,66 @@ buildscript {
...
+sourceSets {
+  main {
+    java.srcDirs += project.tasks.createVersionClass.ext.OUTPUT_DIR
+  }
+}
+
+project.tasks.compileJava.dependsOn(project.tasks.createVersionClass);
+project.tasks.checkstyleMain.dependsOn(project.tasks.createVersionClass);
+
+test {
```
basoko, line 73: Is this needed if we are getting the version from the generated class?

rollbar-java/build.gradle
```diff
@@ -12,12 +12,66 @@ buildscript {
...
+  ext.set("OUTPUT_DIR", [buildDir.getAbsolutePath(), 'src', 'generated', 'main'].join(File.separator))
+
+  outputs.dir(ext.OUTPUT_DIR)
+
+  doLast {
+    def pkg = ["com", "rollbar", "notifier", "provider", "notifier"];
+
+    def pkgName = pkg.join(".");
+    def pkgPath = "${ext.OUTPUT_DIR}${File.separator}" +
+            (['src', 'generated'] + pkg).join(File.separator)
```
basoko, line 42: Do we want this path? Seems that we have two times `src/generated` in the path, just double-checking.

basoko (commented):
This is a nice improvement @diegov ! I left few comments.

bxsx (commented):
Nothing to add from my side. Looks good 👌  Only concerns about @basoko comments.

rollbar-java/build.gradle
diegov: Yup, intellij even warned me about it I thought it was a false positive... :facepalm:

rollbar-java/build.gradle
diegov: There is a test in our suite (versionReturnedShouldMatchManifestVersion below) that is basically making sure our hand-rolled code generation works, this is a way to pass the version to the test that doesn't depend on our code generation. A plain env var is a lot more reliable than our own code generation, so it is good enough as an oracle for the test.

rollbar-java/build.gradle
diegov: Fixed in the latest push

bxsx (approved):

basoko (approved):

diegov:
@basoko requesting re-review since I had accidentally pushed the code review fix to a different branch earlier. Thanks!

bxsx (approved):

PR #323, opened 2025-01-13
[Feat][SDK-347] ANR report

buongarzoni:
## Description of the change

- Update Android min SDK from 16 to 21
- Update Android target/compile SDK from 27 to 30
- Add ANR detectors:
  - Api < 30 uses watchdog
  - Api >= 30 uses historical [ApplicationExitInfo](https://developer.android.com/reference/kotlin/android/app/ActivityManager#gethistoricalprocessexitreasons) 

## How to test
It is necessary to block the main thread
for this, in the Android example, we can add something like 
```java
try {
  Thread.sleep(10000);
  } catch (InterruptedException ignored) {
}
```
in the method clickAction() from the MainActivity
then start clicking in the screen until the alert dialog appears to close the app
when the app is opened again the ANR must be sent.
This feature is enabled by default, and has 2 different implementations depending on witch Android version the user has.

### Watchdog
Monitors the main thread  to detect if it's blocked or unresponsive for a certain period of time (default is 5 seconds), periodically checks if the main thread is still responding (default is 500ms).
This implementation can detect ANRs that have not killed the application.

### HistoricalAnrDetector
This implementation uses the ApplicationExitInfo given by Android OS.
To know if we've already sent an ANR, we save its timestamp (if the user deletes the app's data, the ANRs will be sent again).
This implementation only detects ANRs that have killed the application.

### Next steps
We're not parsing the lock reason right now because I need to modify the RollbarThread, this PR is already to big so I'll add it in another iteration.
I can only go up to API 30 since I need to modify CI so Android runs Java 11, not 8, for API 31 or higher. These changes will also be made in another PR.

## Type of change

- [ ] Maintenance
- [ ] New release

## Related issues

- Fix [SC-]
- Fix #1

### Development

- [ ] Lint rules pass locally

### Code review


rollbar-android/src/main/java/com/rollbar/android/anr/historical/stacktrace/Line.java
```diff
@@ -0,0 +1,11 @@
+package com.rollbar.android.anr.historical.stacktrace;
+
+public final class Line {
+  public int lineno;
```
csaba-ilonka-rollbar, line 4: Sounds like something that should be read-only and hidden?

rollbar-android/src/main/java/com/rollbar/android/anr/historical/stacktrace/StackTraceFactory.java
```diff
@@ -0,0 +1,130 @@
...
+public class StackTraceFactory {
+
+  private static final int STACKTRACE_FRAME_LIMIT = 100;
+
+  public StackTraceFactory(/*todo pass options?*/) {
+  }
+  
+  public List<StackFrame> getStackFrames(
+       final StackTraceElement[] elements, final boolean includeFrames) {
+    List<StackFrame> StackFrames = null;
```
csaba-ilonka-rollbar, line 17: Make sure to stick to a single style convention, preferable something that's already in place.

rollbar-android/src/main/java/com/rollbar/android/anr/historical/stacktrace/StackTraceFactory.java
```diff
@@ -0,0 +1,130 @@
...
+   *
+   * @param className the className
+   * @return true if it is or false otherwise
+   */
+  
+  public Boolean isInApp(final  String className) {
+    if (className == null || className.isEmpty()) {
+      return true;
+    }
+/*
```
csaba-ilonka-rollbar, line 62: Let's not ship code that's commented out.

rollbar-android/src/main/java/com/rollbar/android/anr/historical/stacktrace/ThreadDumpParser.java
```diff
@@ -0,0 +1,327 @@
...
+
+  private final StackTraceFactory stackTraceFactory;
+
+  public ThreadDumpParser(final boolean isBackground) {
+    this.isBackground = isBackground;
+    this.stackTraceFactory = new StackTraceFactory();
+  }
+
+  
+  public List<RollbarThread> parse(final  Lines lines) {
```
csaba-ilonka-rollbar, line 61: Please cover with tests.

rollbar-android/src/main/java/com/rollbar/android/anr/historical/stacktrace/Lines.java
```diff
@@ -0,0 +1,55 @@
+package com.rollbar.android.anr.historical.stacktrace;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.util.ArrayList;
+
+public final class Lines {
```
csaba-ilonka-rollbar, line 7: Why do we need to keep track of the indices? It looks like they’re used to implement some sort of iterator, but couldn’t that be handled by the ArrayList itself?

rollbar-android/src/main/java/com/rollbar/android/anr/historical/stacktrace/LockReason.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class LockReason implements JsonSerializable {
+
+  public static final int LOCKED = 1;
+  public static final int WAITING = 2;
+  public static final int SLEEPING = 4;
+  public static final int BLOCKED = 8;
```
csaba-ilonka-rollbar, line 15: What’s the reason for using numeric constants? Could this be an enum instead?

csaba-ilonka-rollbar: Is it to do with serialization? If so, who is reading the message on the other end?

rollbar-android/src/main/java/com/rollbar/android/anr/historical/stacktrace/LockReason.java
```diff
@@ -0,0 +1,112 @@
...
+  @Override
+  public int hashCode() {
+    return Objects.hash(address);
+  }
+
+  @Override
+  public Object asJson() {
+    Map<String, Object> values = new HashMap<>();
+    values.put(JsonKeys.TYPE, type);
+      if (address != null) {
```
csaba-ilonka-rollbar, line 84: Nitpick - formatting is a bit messed up in this file.

rollbar-android/src/main/java/com/rollbar/android/anr/historical/stacktrace/StackFrame.java
```diff
@@ -0,0 +1,82 @@
...
+  private String function = "";
+  private String module = "";
+  private Integer lineno = 0;
+  private String _package;
+
+  public StackTraceElement toStackTraceElement() {
+    return new StackTraceElement(module, function, filename, lineno);
+  }
+
+  private LockReason lock;
```
csaba-ilonka-rollbar, line 20: Please keep data at the top of the class.

rollbar-android/src/main/java/com/rollbar/android/anr/historical/stacktrace/StackTrace.java
```diff
@@ -0,0 +1,49 @@
...
+public final class StackTrace implements JsonSerializable {
+
+  private final List<StackFrame> frames;
+  private  Boolean snapshot;
+
+  public StackTrace(final  List<StackFrame> frames) {
+    this.frames = frames;
+  }
+
+  public StackTraceElement[] getStackTraceElements() {
```
csaba-ilonka-rollbar, line 18: Could this return a `List< StackTraceElement>` instead of an array? it would simplify the method.

rollbar-android/src/main/java/com/rollbar/android/anr/historical/stacktrace/ThreadParser.java
csaba-ilonka-rollbar, line 1: Needs tests. 😢

csaba-ilonka-rollbar (commented):
Requires significantly more test coverage.

rollbar-android/src/main/java/com/rollbar/android/anr/historical/stacktrace/StackTrace.java
buongarzoni: This is because we also use the [thread stacktrace](https://github.com/rollbar/rollbar-java/blob/b88ad18b0e8bfba394e63e64560e2ae462fd4325/rollbar-java/src/main/java/com/rollbar/notifier/util/BodyFactory.java#L86) which is an array, but I simplified the method in this commit caf072fa

rollbar-android/src/main/java/com/rollbar/android/anr/historical/stacktrace/LockReason.java
buongarzoni: I removed the LockReason for now, I think it would be better to add it in another PR

rollbar-android/src/main/java/com/rollbar/android/anr/historical/stacktrace/LockReason.java
buongarzoni: I removed the LockReason for now, I think it would be better to add it in another PR

rollbar-android/src/main/java/com/rollbar/android/anr/historical/stacktrace/Lines.java
buongarzoni: I deleted this class

PR #374, opened 2026-05-26
[Feat] [SDK-399] Add java agent for network telemetry events

buongarzoni:
## Description of the change

#### Add Java agent for automatic network telemetry capture                                                                                                                                                                            
                                                                                                                                                                                                                                         
  Auto-instruments all major HTTP clients via `-javaagent:` using ByteBuddy, capturing 4xx/5xx responses as Rollbar telemetry events with **no changes at HTTP call sites**.

Scope of "no code changes": request code is never touched — no wrappers, no interceptors, no per-call bookkeeping, and nothing to remember when a new HTTP call is added. Setup is a one-time wiring step: the agent JAR on the application classpath, and `.telemetryEventTracker(RollbarAgent.getTelemetryTracker())` on the config builder.

That wiring is not automatic by design of the current SDK: `ConfigBuilder.build()` installs its default `RollbarTelemetryEventTracker` whenever `telemetryEventTracker(...)` was not called, and there is no global registry or `ServiceLoader` hook an agent could claim instead. Making the agent self-installing would require a change to `rollbar-java` and is tracked separately.

  > [!CAUTION]
  > This module targets JVM-based applications only. Android is not supported — ART does not implement
  > the `java.lang.instrument` API required by Java agents. Android users should use the existing
  > `rollbar-android` module instead.

  The acceptance criteria on the Shortcut story need the same narrowing — "zero application code changes" → "no changes at HTTP call sites; one-time tracker wiring at init". 

####  What's included:
  - New rollbar-java-agent module — shadow JAR with ByteBuddy bundled and relocated
  - Instruments HttpURLConnection, java.net.http.HttpClient, Apache HC 4.x and 5.x
  - URL sanitization (strips credentials, query params, fragment before recording)
  - Deduplication via WeakHashMap to handle re-entrant getResponseCode() calls and dual-advice firing on HttpClient
  - Integration tests using WireMock 3.x for each instrumented client
  - README with installation and manual testing guide

  Usage:
```
  -javaagent:/path/to/rollbar-java-agent.jar
```

```java
  Rollbar.init(withAccessToken("...")
      .telemetryEventTracker(RollbarAgent.getTelemetryTracker())
      .build());
```

## Type of change

- [ ] Maintenance
- [ ] New release

## Related issues

- Fix [SC-]
- Fix #1

## Checklists

### Development

- [ ] Lint rules pass locally

### Code review

### DIFF
--- a/rollbar-java-agent/README.md
+++ b/rollbar-java-agent/README.md
@@ -15,7 +15,7 @@ It works by attaching to the JVM at startup via `-javaagent:` and using ByteBudd
 
 Only 4xx and 5xx responses are recorded, along with requests that fail before a response arrives (connection refused, DNS failure, timeout). Successful requests (< 400) produce no telemetry.
 
-> **Apache HC4/HC5 note:** `execute(request, responseHandler)` overloads route through a separate internal call chain and are not instrumented. Use the standard `execute(request)` / `execute(request, context)` overloads to get telemetry.
+**Apache HC4/HC5:** every `execute(...)` overload is covered — the request-only forms, the target-host forms (`execute(HttpHost, request)`), and the response-handler forms. The agent instruments the protected `doExecute(HttpHost, request, context)` method that all of them converge on, rather than any individual `execute()` overload, so no dispatch path is missed. Requests issued through a target-host overload carry only a path, so the agent rejoins the host from the `HttpHost` argument to record a complete URL.
 
 ### HttpURLConnection entry points
 
--- a/rollbar-java-agent/README.md
+++ b/rollbar-java-agent/README.md
@@ -1,8 +1,8 @@
 # Rollbar Java Agent
 
-A zero-code-change Java instrumentation agent that automatically captures HTTP network errors (4xx and 5xx responses) as Rollbar telemetry events.
+A Java instrumentation agent that automatically captures HTTP network errors (4xx and 5xx responses) as Rollbar telemetry events, with no changes at your HTTP call sites.
 
-It works by attaching to the JVM at startup via `-javaagent:` and using ByteBuddy to intercept HTTP calls across all major clients — no library dependencies or code changes are needed in your application.
+It works by attaching to the JVM at startup via `-javaagent:` and using ByteBuddy to intercept HTTP calls across all major clients. Your request code stays exactly as it is, and you add no HTTP-related library dependencies. Setup is a one-time wiring step in your Rollbar configuration — see [What "no code changes" means here](#what-no-code-changes-means-here).
 
 ## Instrumented HTTP clients
 
@@ -41,6 +41,22 @@ application uses and never pins or shadows your chosen SDK version.
 
 ## Installation
 
+### What "no code changes" means here
+
+All four steps below are **required**. Steps 3 and 4 touch your application once, at setup:
+
+- **What you never change:** your HTTP call sites. Every request through `HttpURLConnection`,
+  `java.net.http.HttpClient`, or Apache HC 4.x/5.x is instrumented as written — no wrappers, no
+  interceptors, no per-call bookkeeping, and nothing to remember when you add the next HTTP call.
+- **What you change once:** the agent JAR goes on your application classpath (step 3), and your
+  `Rollbar.init(...)` passes `RollbarAgent.getTelemetryTracker()` to the config builder (step 4).
+
+That wiring cannot be made automatic today. `ConfigBuilder.build()` installs its default
+`RollbarTelemetryEventTracker` whenever `telemetryEventTracker(...)` was not called, and the SDK
+exposes no global registry or `ServiceLoader` hook that an agent could claim instead — so the
+tracker has to be handed to the builder by the application. Skipping step 4 is silent: the agent
+still records events, but into a store nothing ever reads (see [Behavior](#behavior)).
+
 ### 1. Build the agent JAR
 
 ```bash
@@ -78,9 +94,11 @@ jvmArgs("-javaagent:/path/to/rollbar-java-agent-<version>.jar")
 JAVA_TOOL_OPTIONS="-javaagent:/path/to/rollbar-java-agent-<version>.jar"
 ```
 
-### 3. Also add the JAR to your classpath
+### 3. Also add the JAR to your classpath (required, for compile time)
 
-The agent JAR must also be available on the regular application classpath so that your code can call `RollbarAgent.getTelemetryTracker()`:
+At runtime the JVM already appends a `-javaagent:` JAR to the system class path, so this step is not
+about making the agent load. It is about **compiling** step 4: your build needs the JAR as an
+ordinary dependency to resolve the `RollbarAgent` symbol.
 
 **Gradle:**
 ```kotlin
@@ -98,7 +116,7 @@ dependencies {
 </dependency>
 ```
 
-### 4. Wire into your Rollbar configuration
+### 4. Wire into your Rollbar configuration (required)
 
 ```java
 import com.rollbar.agent.RollbarAgent;
@@ -114,7 +132,7 @@ Rollbar rollbar = Rollbar.init(
 );
 ```
 
-That's it. All HTTP calls your application makes from that point on will automatically produce telemetry events in the Rollbar error report for any 4xx or 5xx response.
+That's the last application change you make. From here on, every HTTP call — including ones you add later — automatically produces a telemetry event in the Rollbar error report for any 4xx or 5xx response, with no further code changes.
 
 ## Behavior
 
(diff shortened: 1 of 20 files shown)


buongarzoni:
@claude review

buongarzoni:
@claude review

rollbar-java-agent/build.gradle.kts
```diff
@@ -0,0 +1,56 @@
...
+            "Premain-Class" to "com.rollbar.agent.RollbarAgent",
+            "Agent-Class" to "com.rollbar.agent.RollbarAgent",
+            "Can-Redefine-Classes" to "true",
+            "Can-Retransform-Classes" to "true"
+        )
+    }
+}
+
+tasks.shadowJar {
+    archiveClassifier.set("")
```
brianr: [P1] Avoid letting the plain jar replace the shaded agent

With the classifier set to empty while the normal `jar` task remains enabled (and `test` depends on both tasks), `jar` and `shadowJar` write the same `rollbar-java-agent-<version>.jar`. In any build/publish/test graph where `jar` runs after `shadowJar`, the path used by `-javaagent` is a thin jar without relocated ByteBuddy, so the agent fails to start; give the shaded jar its own artifact or disable/remap the plain jar.

rollbar-java-agent/src/main/java/com/rollbar/agent/instrumentation/ApacheHttpClient4Instrumentation.java
```diff
@@ -0,0 +1,113 @@
...
+   *
+   * <p>Does nothing if {@code org.apache.http.impl.client.CloseableHttpClient} is not available.
+   */
+  public static void installIfAvailable(AgentBuilder builder, Instrumentation inst) {
+    // Use a resource check rather than Class.forName: execute(HttpUriRequest, HttpContext) is a
+    // concrete method defined in CloseableHttpClient itself, so loading that class before the
+    // ByteBuddy transformer is installed prevents the method from ever being instrumented (no
+    // RedefinitionStrategy is configured). getResource checks classpath availability without
+    // triggering class loading.
+    if (ClassLoader.getSystemClassLoader().getResource(
```
brianr: [P2] Install Apache transformers without system-loader gating

When Apache HttpClient is provided by a child/application classloader (for example servlet containers, Gradle test isolation, or module layers), `ClassLoader.getSystemClassLoader().getResource(...)` can return null during premain even though ByteBuddy would later see `CloseableHttpClient` as it is loaded. This returns before installing the transformer, so HC4 requests in those deployments produce no telemetry; the analogous HC5 check needs the same treatment.

rollbar-java-agent/src/main/java/com/rollbar/agent/instrumentation/HttpUrlConnectionInstrumentation.java
```diff
@@ -0,0 +1,83 @@
...
+   * Instruments {@code java.net.HttpURLConnection.getResponseCode()} to record 4xx/5xx responses.
+   *
+   * <p>Targets the declaring class directly to capture the method regardless of concrete subtype.
+   */
+  public static void install(AgentBuilder builder, Instrumentation inst) {
+    builder
+        .type(ElementMatchers.named("java.net.HttpURLConnection"))
+        .transform((b, typeDescription, classLoader, module, protectionDomain) ->
+            b.visit(Advice.to(GetResponseCodeAdvice.class)
+                .on(ElementMatchers.named("getResponseCode")))
```
brianr, line 38: [P2] Capture HttpURLConnection requests that skip getResponseCode

For `HttpURLConnection` callers that trigger the request with `getInputStream()` or `getErrorStream()` and never call `getResponseCode()`, this is the only advised method, so a 4xx/5xx response (or the `IOException` thrown by `getInputStream()` on 4xx) is never recorded. This leaves a common `HttpURLConnection` usage path outside the promised automatic network-error capture.

rollbar-java-agent/src/main/java/com/rollbar/agent/UrlSanitizer.java
```diff
@@ -0,0 +1,33 @@
...
+          uri.getScheme(),
+          null,
+          uri.getHost(),
+          uri.getPort(),
+          uri.getPath(),
+          null,
+          null
+      ).toString();
+    } catch (URISyntaxException e) {
+      return rawUrl;
```
brianr: [P2] Do not return unsanitized URLs on parse failures

If a URL accepted by `java.net.URL` but rejected by `URI` reaches this catch, such as one with an unescaped space or bad percent escape plus `?token=...`, returning `rawUrl` records the original query/userinfo even though this sanitizer is the last line of defense before telemetry. The fallback should still strip credentials, query, and fragment conservatively instead of failing open.

brianr (commented):
Posting review findings from the local review.

rollbar-java-agent/src/main/java/com/rollbar/agent/UrlSanitizer.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+    try {
+      URI uri = new URI(rawUrl);
+      // Use getAuthority() rather than getHost(): URI parses authorities that fail RFC 2396
+      // server-based grammar (e.g. underscores in Kubernetes/AD internal DNS) in registry-based
+      // mode, where getHost() returns null and the host would be silently dropped. Strip userinfo
+      // from the authority manually.
+      String authority = uri.getAuthority();
+      if (authority != null) {
+        int at = authority.indexOf('@');
```
brianr: [P1] Preserve encoded credentials while locating the userinfo delimiter

`URI.getAuthority()` returns the **decoded** authority. For `https://user:p%40ss@example.com/path`, it yields `user:p@ss@example.com`; `indexOf('@')` then keeps `ss@example.com`, so part of the password is recorded as new userinfo. Please operate on `getRawAuthority()` / `getRawUserInfo()` (or otherwise locate the delimiter in the raw authority) before reconstructing the sanitized URL, and add a regression test with percent-encoded `@` in userinfo.

Oracle documents the decoding behavior here: https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/net/URI.html#getAuthority()

rollbar-java-agent/build.gradle.kts
```diff
@@ -0,0 +1,65 @@
+plugins {
+    `java-library`
+    id("com.github.johnrengelman.shadow") version "8.1.1"
+}
+
+dependencies {
+    implementation("net.bytebuddy:byte-buddy:1.14.18")
```
brianr: [P1] Support the advertised JVM range

Byte Buddy 1.14.18 supports class files only through Java 23 without experimental mode; Java 24 support starts at 1.15.4 and Java 25+ at 1.17.0. The README promises Java 11 or higher, but CI exercises only 11 and 17, so the agent can silently fail to transform HTTP classes on current Java 24/25/26 runtimes. Please upgrade Byte Buddy and add current-LTS/current-JDK coverage, or explicitly narrow the supported range.

Compatibility table: https://github.com/raphw/byte-buddy#java-version-compatibility

rollbar-java-agent/build.gradle.kts
```diff
@@ -0,0 +1,65 @@
...
+    implementation("net.bytebuddy:byte-buddy-agent:1.14.18")
+    api(project(":rollbar-api"))
+    implementation(project(":rollbar-java"))
```
brianr, line 33: [P1] Keep Rollbar SDK classes out of the shaded agent

`shadowJar` merges `runtimeClasspath` by default, so this `implementation` dependency and the `rollbar-api` `api` dependency are embedded unrelocated in the fat JAR (along with transitive dependencies such as SLF4J); only Byte Buddy is relocated below. The README also puts the agent alongside the application's Rollbar SDK, creating duplicate `com.rollbar.*` classes that can pin/override another SDK version or split class identity in containers. Please shade only agent-private dependencies (for example via a dedicated Byte Buddy configuration) and leave the Rollbar API/SDK as ordinary external dependencies.

Shadow's default dependency behavior: https://gradleup.com/shadow/configuration/dependencies/

rollbar-java-agent/README.md
```diff
@@ -0,0 +1,202 @@
+# Rollbar Java Agent
+
+A zero-code-change Java instrumentation agent that automatically captures HTTP network errors (4xx and 5xx responses) as Rollbar telemetry events.
```
brianr: [P1] Align the implementation with the zero-code-change claim

The instrumentation removes changes at HTTP call sites, but installation still requires adding the agent as an application dependency and changing `Rollbar.init(...)` to install `RollbarAgent.getTelemetryTracker()` (steps 3–4). The behavior table also says that without this wiring events accumulate but are not sent. If SDK-399 requires genuinely zero application-code changes, the agent needs to integrate its tracker automatically; otherwise this claim should be narrowed to "no HTTP client call-site changes" and the stated problem/acceptance criteria updated.

brianr (changes requested):
Requesting changes on the current head. The inline comments cover four release blockers. Two existing review threads also still apply, so I have not duplicated them: the agent still records Rollbar's own `SyncSender` failures, and `NetworkEventBridge.composeUrl()` still treats `://` inside a relative URI's query/path as an absolute URI. The former thread is marked resolved even though no suppression guard is present at this head.

rollbar-java-agent/build.gradle.kts
waltjones: This has been addressed.

waltjones:
@buongarzoni thoughts on this? Claude was able to reproduce this issue, and claims the majority of the sample apps would be affected by it.

The JVM refuses to start unless the Rollbar SDK is on the system classpath.

```
Exception in thread "main" java.lang.NoClassDefFoundError: com/rollbar/notifier/telemetry/TelemetryEventTracker
      at java.base/java.lang.Class.getDeclaredMethods0(Native Method)
      at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndStartAgent(InstrumentationImpl.java:452)
      at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndCallPremain(InstrumentationImpl.java:503)
...
FATAL ERROR in native method: processing of -javaagent failed, processJavaStart failed
```

This is only when the new agent is configured.

PR #376, opened 2026-06-05
fix(logback): restrict recursive logger suppression to Rollbar package

Montana:
Only suppress events emitted by the `com.rollbar` logger or one of its package descendants.

The previous recursive logging check used a broad prefix match against `com.rollbar`. As a result, unrelated application loggers with names that happened to begin with the same characters were incorrectly treated as internal Rollbar loggers and silently skipped. For example, a logger such as `com.rollbarista.checkout` would never be forwarded to Rollbar even though it is not part of the Rollbar package.

So this updates the check so it only matches the exact `com.rollbar` logger or names beginning with the `com.rollbar.` package boundary. This preserves the recursive logging protection for Rollbar’s own internal logs while allowing similarly named application packages to be reported normally.

### DIFF
--- a/rollbar-logback/src/main/java/com/rollbar/logback/RollbarAppender.java
+++ b/rollbar-logback/src/main/java/com/rollbar/logback/RollbarAppender.java
@@ -108,12 +108,12 @@ public class RollbarAppender extends AppenderBase<ILoggingEvent> {
     super.start();
   }
 
-  @Override
-  protected void append(ILoggingEvent event) {
-    if (isRollbarLogger(event.getLoggerName())) {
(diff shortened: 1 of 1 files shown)


Montana (commented):
Cleaned up the implementation and removed the accidental extra formatting changes so the final diff is focused only on the recursive logger suppression fix. 

The branch now includes a few incremental cleanup commits from editing through the GitHub interface, but the resulting code is in the intended state. The maintainer can squash the commits during merge to keep the history clean.

buongarzoni (approved):

brianr (approved):

PR #377, opened 2026-07-13
[Feat][SDK-601] Add built in scrubbing

buongarzoni:
## Description of the change

Adds a scrubbing layer that runs on every payload, so sensitive data is redacted by default rather than only when a user wires up their own transformer.

### What it does

  `ScrubDataTransformer` runs after any user-supplied transformer and cannot be
  bypassed by replacing it. It redacts values with `***` in:

  - **Fields whose key names a secret**, with no configuration: `password`, `passwd`,
    `secret`, `token`, `authorization`, `authentication`, `^auth$`, `apikey`/`api_key`/
    `api-key`. Case-insensitive, matched anywhere in the key (so `token` covers
    `access_token` and `csrfToken`); `auth` is anchored so `author` is left alone.
 - **Request headers** — via a built-in deny-list (`Authorization`, `Cookie`,
    `Set-Cookie`, `X-Api-Key`, `X-Auth-Token`, `X-Access-Token`, `X-Secret`,
    `Proxy-Authorization`, `WWW-Authenticate`). Always on, no config needed.
- **Every slot a request is serialized into**: `request.get`, `request.post`,
    `request.params`, `request.metadata`, `request.query_string`, custom data and
    `Frame.locals`, including the copies carried by `body.threads`. Nested maps,
    collections and arrays are traversed to a depth of 8, preserving shape.

  `Request.url` is sanitized unconditionally by `DefaultUrlSanitizer`, which strips
  userinfo, query string, and fragment.

Not covered: `request.body`, a raw string the notifier cannot parse.

### Config

  ```java
  ConfigBuilder.withAccessToken(TOKEN)
      .redactedKeys(Arrays.asList("ssn", "pin"))  // case-insensitive regex, additive
      .useDefaultRedactedKeys(false)              // optional: match only my keys
      .urlSanitizer(myCustomSanitizer)            // optional
      .build();
  ```

  Added to both the core and reactive-streams builders. useDefaultRedactedKeys
  defaults to true on CommonConfig, so third-party Config implementations get the
  built-in redaction too.

  >  [!WARNING]
  >  Behavior change. Two things arrive differently in Rollbar after this PR:
  >  1. Fields matching the built-in key list are redacted, including keys that are not
  >  strictly secrets — substring matching means tokenCount is redacted too. Set
  >  useDefaultRedactedKeys(false) if that is too broad for your payloads.
  >  2. Query strings are stripped from request.url. Users who relied on seeing them
  >  need a custom urlSanitizer.

### Performance

  Keys without regex syntax — the whole built-in list, and most user keys — are matched
  with a lowercase pass and a substring search rather than a Matcher per key per
  pattern. On a ~100-key payload that is 6.3 µs and 7.5 KB allocated, down from 28.7 µs
  and 171 KB. Regex semantics for redactedKeys are unchanged; keys that do carry regex
  syntax still run through Pattern.

### okhttp

RollbarOkHttpInterceptor now delegates its default URL sanitization to the
shared DefaultUrlSanitizer so the okhttp and notifier paths can't drift apart,
and gains a withSharedUrlSanitizer(recorder, sanitizer) factory to reuse the
sanitizer from a notifier config.

The sanitizer types live in rollbar-api (com.rollbar.api.scrubbing), not
rollbar-java, so rollbar-okhttp keeps its existing lightweight dependency.
The factory is static rather than a constructor overload — UrlSanitizer and
StringUrlSanitizer are both functional interfaces, so an overload would make a
lambda argument ambiguous and break existing callers at compile time.

### Testing

  - Unit coverage of the key list, the opt-out, nesting, collections/arrays, query-string
  encoding, and literal-vs-regex matching equivalence.
  - End-to-end tests against the JSON WireMock receives, including one that sends a
  /login?password=hunter2 request through a bare ConfigBuilder and asserts the
  secret appears nowhere in the payload.
  - Docs in SCRUBBING.md
  
## Type of change

- [ ] Maintenance
- [ ] New release

## Related issues

- Fix [SC-]
- Fix #1

### Development

- [ ] Lint rules pass locally

### Code review


buongarzoni:
@claude review

rollbar-java/src/main/java/com/rollbar/notifier/scrubbing/ScrubDataTransformer.java
```diff
@@ -0,0 +1,435 @@
...
+    Map<String, Object> result = null;
+    for (Map.Entry<String, Object> entry : map.entrySet()) {
+      String key = entry.getKey();
+      Object value = entry.getValue();
+      if (matchesAny(key, patterns)) {
+        if (result == null) {
+          result = new HashMap<>(map);
+        }
+        result.put(key, SCRUBBED_VALUE);
+      } else if (value instanceof Map && depth < MAX_SCRUB_DEPTH) {
```
brianr: [P1] Traverse collections and arrays when scrubbing nested data

This only descends into values that are directly `Map` instances. Valid payloads such as `custom = {"users": [{"password": "hunter2"}]}` therefore ship the configured secret unchanged. The serializer supports both `Collection` and `Object[]`, and the same bypass applies to nested values in `Request.post`, `Request.metadata`, and `Frame.locals`. Please recursively scrub maps contained in collections/arrays under the same depth cap, preserve the surrounding shape, and add regression tests for both forms.

rollbar-java/src/main/java/com/rollbar/notifier/scrubbing/ScrubDataTransformer.java
```diff
@@ -0,0 +1,435 @@
...
+      return null;
+    }
+
+    Request originalRequest = data.getRequest();
+    Map<String, Object> originalCustom = data.getCustom();
+    Body originalBody = data.getBody();
+
+    Request scrubbedRequest = scrubRequest(originalRequest);
+    Map<String, Object> scrubbedCustom = scrubObjectMap(originalCustom, fieldPatterns, 0);
+    Body scrubbedBody = scrubBody(originalBody);
```
brianr, line 139: [P1] Sanitize URLs in network telemetry events

`RollbarBase.recordNetworkEventFor(...)` passes its URL directly to the tracker, which stores it unchanged. Those events are later carried in `Body.telemetryEvents`, but `scrubBody()` only rebuilds trace content and thread traces. A caller can therefore record `https://user:pass@example.com/path?token=secret` and ship both userinfo and query string despite the built-in scrubber. Please apply the configured sanitizer before recording network telemetry, or rebuild network telemetry events here, and cover the public notifier API with a regression test.

brianr (changes requested):
The always-last transformer and copy-on-write implementation are a good foundation, and the current head fixes the earlier routing-param, metadata, encoded-query-key, and JVMTI thread-local gaps. I am requesting changes for the two remaining payload paths called out inline. Before release, this behavior change also needs user-facing migration documentation plus an integration test proving transformer ordering/reconfiguration (including the reactive builder) and the shared OkHttp sanitizer path.

rollbar-java/src/main/java/com/rollbar/notifier/config/ConfigBuilder.java
```diff
@@ -659,6 +696,10 @@ private static class ConfigImpl implements Config {
       this.compressPayload = builder.compressPayload;
       this.maximumTelemetryData = builder.maximumTelemetryData;
       this.telemetryEventTracker = builder.telemetryEventTracker;
+      this.redactedKeys = builder.redactedKeys != null
```
brianr, line 720: [P1] Seed field scrubbing with a meaningful default list

When `redactedKeys` is unset, this stores an empty list, so `ScrubDataTransformer` leaves GET/POST parameters, `request.query_string`, custom data, metadata, and `Frame.locals` untouched. Stripping the query from `request.url` does not close that gap: the web request provider serializes parsed GET parameters and the raw query string separately, so `/login?password=hunter2` still sends `hunter2` under both `request.get.password` and `request.query_string` with the default configuration.

That means the default behavior only protects the built-in header deny-list and URL strings; it does not satisfy the stated goal that sensitive data is redacted by default. Please seed this with an additive built-in field list (for example password/passwd, secret, token/access_token, auth/authentication/authorization), ideally with an explicit override mechanism, and add a no-configuration end-to-end test proving the secret is absent from every relevant request representation.

brianr (commented):
One additional security issue on the current head.

waltjones:
Claude analysis

## Status of existing review comments

All six inline comments are addressed in the current head, including the three brianr threads still marked unresolved:
```
┌──────────────────────────────────────────────┬──────────────────┬───────────────────────────────────────────────────────────────┐
│                   Comment                    │     Fixed by     │                           Verified                            │
├──────────────────────────────────────────────┼──────────────────┼───────────────────────────────────────────────────────────────┤
│ codex P1 — locals in body.threads            │ 82fb70b          │ scrubThreads/scrubThread walk Body.getRollbarThreads()        │
├──────────────────────────────────────────────┼──────────────────┼───────────────────────────────────────────────────────────────┤
│ codex P2 — percent-encoded query keys        │ 8c06fd1          │ decodeParamName, matched against raw and decoded              │
├──────────────────────────────────────────────┼──────────────────┼───────────────────────────────────────────────────────────────┤
│ claude — Request.params / metadata / threads │ a09f485, 82fb70b │ all three in scrubRequest/scrubBody                           │
├──────────────────────────────────────────────┼──────────────────┼───────────────────────────────────────────────────────────────┤
│ brianr P1 — collections & arrays             │ 57399ec          │ scrubCollection/scrubArray, depth-capped, shape preserved     │
├──────────────────────────────────────────────┼──────────────────┼───────────────────────────────────────────────────────────────┤
│ brianr P1 — network telemetry URLs           │ 5482e72          │ sanitized at record time in RollbarBase.recordNetworkEventFor │
├──────────────────────────────────────────────┼──────────────────┼───────────────────────────────────────────────────────────────┤
│ brianr P1 — default key list                 │ 1fc3ed6, f016ac2 │ DEFAULT_REDACTED_KEYS + useDefaultRedactedKeys                │
└──────────────────────────────────────────────┴──────────────────┴───────────────────────────────────────────────────────────────┘
```
brianr's review body also asked for migration docs and integration tests for ordering/reconfiguration/okhttp. Docs ✓ (SCRUBBING.md "Migrating"), okhttp ✓, ordering + reconfiguration ✓ in ScrubbingITest — but ScrubbingReactiveITest covers ordering only, no reconfiguration test on the reactive builder, which is the half brianr specifically called out.

The core design is right: always-last transformer that can't be replaced, copy-on-write so unchanged payloads keep identity, eager Pattern.compile so a bad regex fails at configure time. KeyMatcher's literal fast path is a genuinely good call and its Unicode-vs-ASCII folding divergence is documented and errs toward redacting more.

## New findings

1. Message.metadata is not scrubbed — ScrubDataTransformer.java:210 (scrubBodyContent)
Only Trace and TraceChain are handled; Message falls through untouched. Message.metadata is a Map<String,Object> flattened straight into body.message.* (Message.java:48-53) — structurally identical to Request.metadata, which this PR does scrub. Nothing in-tree populates it (BodyFactory:179 sets body only), so it's reachable only via a user Transformer — but that is exactly the argument that was accepted for scrubbing Request.metadata. Two lines in scrubBodyContent.

2. Telemetry event bodies are never scrubbed, and the docs don't say so — SCRUBBING.md:45
TelemetryEvent.body is a Map<String,String> that scrubBody doesn't touch. Only the network URL is protected, and only because it's sanitized at record time. recordLogEventFor / recordManualEventFor / recordNavigationEventFor bodies ship verbatim. That matters on Android: LogcatTelemetryCapture:140 records raw logcat lines from the whole app UID including third-party libraries as telemetry. SCRUBBING.md's "Not covered" section lists only request.body. It should also list telemetry bodies, body.message metadata, Data.person and Data.server — a scrubbing doc that undersells its gaps is worse than one that has gaps.

3. SCRUBBING.md:22 overstates coverage — "Fields whose key names a secret, wherever they appear in the payload." It's a specific set of slots, correctly enumerated 40 lines later. Drop the "wherever they appear".

4. An empty string in redactedKeys silently redacts the entire payload — ScrubDataTransformer.java:~560 (KeyMatcher.of)
isLiteral("") is true → "" lands in contains → lower.contains("") always matches. (Pattern.compile("").find() does the same, so it's inherent to the regex contract rather than a fast-path bug — but it's now reachable from a config file with a blank line or trailing comma, and the failure is silent and total.) Skip null/blank entries in KeyMatcher.of.

waltjones:
@buongarzoni Let me know what you think about 1 and 2 above. Should those sections be scrubbed in this PR, in a follow-up, or is there a reason to skip them?

waltjones (approved):
👍 All looks good.

PR #379, opened 2026-07-30
[Fix][issue-227] Native agent 10x start up

### LINKED ISSUE #227: Native agent adds 10x startup time to Spring Boot app
I added a sample project to reproduce it at https://github.com/phillipuniverse/rollbar-native-test. Without the agent, startup time is ~2s. With the agent, startup time is ~20s. The project uses the `.dylib` and `.so` files that I used to validate with, both with the same results (although Linux was slightly faster). I build the `.dylib` and `.so` files from the master branch just to make sure I had all the changes.

In this example it isn't so bad, but it takes a different Spring Boot project (that has Spring Data, Rabbit, Mongo, JPA, etc etc) go from ~15s startup time to ~400s which makes the whole native agent a non-starter.

I will say though, once it starts up I get locals in my Rollbar traces which is why I would really like this to work!!

image

Might be relevant, but when I turn on debug logging with `ROLLBAR_LOG=debug` I get this:

```console
 INFO  rollbar_java_agent > Agent load begin
 INFO  rollbar_java_agent > Agent load complete success
 DEBUG rollbar_java_agent > JNI call failed: 'com/rollbar/jvmti/ThrowableCache class not found: com/rollbar/jvmti/ThrowableCache'
 DEBUG rollbar_java_agent > JNI call failed: 'com/rollbar/jvmti/ThrowableCache class not found: com/rollbar/jvmti/ThrowableCache'
 DEBUG rollbar_java_agent > JNI call failed: 'com/rollbar/jvmti/ThrowableCache class not found: com/rollbar/jvmti/ThrowableCache'
 DEBUG rollbar_java_agent > JNI call failed: 'com/rollbar/jvmti/ThrowableCache class not found: com/rollbar/jvmti/ThrowableCache'
 DEBUG rollbar_java_agent > JNI call failed: 'com/rollbar/jvmti/ThrowableCache class not found: com/rollbar/jvmti/ThrowableCache'
 DEBUG rollbar_java_agent > JNI call failed: 'com/rollbar/jvmti/ThrowableCache class not found: com/rollbar/jvmti/ThrowableCache'
 DEBUG rollbar_java_agent > JNI call failed: 'com/rollbar/jvmti/ThrowableCache class not found: com/rollbar/jvmti/ThrowableCache'
...
...
```

The `ThrowableCache class not found` is emitted an insane amount of times. If I were to bet, it seems to indicate that the cache is not being used by the native agent, which contributes to the super slow startup times??

buongarzoni:
## Description of the change

The JVMTI native agent made JVM startup 2–10x slower (reporter saw 15s → 400s on a Spring Boot app). Three independent causes:

  1. The Exception event was armed at Agent_OnLoad for all threads, so every exception thrown during VM and framework bootstrap entered the callback, tens of thousands of them, before ThrowableCache was even loadable.
  2. FindClass could not resolve ThrowableCache. Called from a JVMTI callback with no Java frame, it resolves against the system classloader, so inside a Spring Boot fat jar it failed permanently, the agent was pure overhead and captured nothing. Each failure also ran a Java upcall to build the error message, at every log level.
  3. Nothing was cached, and locals were captured for every frame, including the 100+ framework frames nobody wants locals from.

  Changes
  - Activation is gated on a ClassPrepare callback for com.rollbar.jvmti.ThrowableCache; the supplied jclass is cached as a global ref, so FindClass is never used for Rollbar classes again. This fixes fat ar and container classloaders.
  - All classes, method IDs, and boxing helpers resolved once instead of per-event and per-frame.
  - Native app package pre filter (memoized per jmethodID) and locals capture restricted to app frames. Frame array stays 1:1 with the stack so BodyFactory.frames() is unaffected.
  - Re entrancy guard, PushLocalFrame/PopLocalFrame, frame cap.
  - ThrowableCache.appPackages is now copy-on-write volatile (was an unsynchronized HashSet read from arbitrary throwing threads), plus a version counter the agent reads cheaply. Older agents/jars degrade gracefully in both directions.
  - GetEnv requests JVMTI 1.2 rather than the build JDK's version, a JDK 21 build previously refused to load on JDK 17.
  - Separately: the JSON serializer no longer throws when a captured local's toString() returns null or throws. This surfaced only once locals capture started working.

## Type of change

- [ ] Maintenance
- [ ] New release

## Related issues

- Fix [#1](https://github.com/rollbar/rollbar-java/issues/227)

### Development

- [ ] Lint rules pass locally

### Code review

### DIFF
--- a/native-agent/Cargo.toml
+++ b/native-agent/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Andrew Weiss <wvvwwvw@gmail.com>"]
 build = "build.rs"
 
 [lib]
-crate_type = [ "cdylib" ]
+crate-type = [ "cdylib" ]
 
 [dependencies]
 jni = "0.21.1"
--- a/native-agent/README.md
+++ b/native-agent/README.md
@@ -42,14 +42,40 @@ new Rollbar(withAccessToken("ACCESS-TOKEN")
 Once you have the agent setup and `rollbar-java` configured, `rollbar-java` will attribute the exceptions
 using the agent as well as send back unhandled exceptions if configured.
 
+Two things are required, and the agent captures nothing if either is missing:
+
+* **`appPackages` must be configured.** It is not an optional filter — it is the switch that
+  turns capture on. With no app packages the agent stays dormant.
+* **Your classes must be compiled with debug information** (`javac -g`, which Gradle and Maven
+  enable by default for `debug = true`). Local variable names live in the `LocalVariableTable`
+  attribute; without `-g` the agent has nothing to read and frames come back with no locals.
+
+Note also that the cache is keyed per thread: locals are only available if the throwable is
+reported on the same thread that threw it. Handing an exception to another thread (an executor,
+an async error handler) before reporting it loses the locals.
+
+## How the agent activates
+
+The agent does nothing until `com.rollbar.jvmti.ThrowableCache` is loaded. It watches for that
+class via a JVMTI `ClassPrepare` callback and only then enables exception capture, so an
+application that never initialises Rollbar — and the whole of JVM and framework startup before
+Rollbar *is* initialised — pays essentially nothing.
+
+This also means the agent resolves `ThrowableCache` through whichever classloader actually
+loaded it. 
+
 Regardless of your JVM language of choice, at some level there will be an invocation of the JVM and
 therefore there is a configuration option to pass arguments directly to the JVM.
 
 ## Getting the agent library
 
-We will attempt to distribute via the releases page pre-built versions of the agent library for
-various architectures. However, if you are running in an environment where one of these libraries
-does not work, then you can build your own as long as you can install the Rust toolchain.
+No pre-built binaries have been published since 1.4.1, and nothing in CI builds this library, so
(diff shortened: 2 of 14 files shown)


waltjones (dismissed):
Approving. These look interesting from the Claude analysis.

1. exceptions.rs:277 — the agent now captures locals for the first time, which exposes that ToReflectedMethod returns a Constructor for <init>/<clinit>, whose getName() is the class name. BodyFactory's name-based resync breaks on any stack with a constructor frame and drops locals for everything above it — defeating the PR's central alignment invariant. All three new alignment tests pass because none include an <init> frame.
2. exceptions.rs:317 — jfloat reaches the C-variadic CallStaticObjectMethod through a generic T, so rustc's E0617 never fires and no float→double promotion happens. HotSpot reads 8 bytes where 4 were written; every captured float is garbage. class_for_name promotes its jboolean with a comment about exactly this rule.
3. exceptions.rs:255 — one failing GetLocalXxx unwinds the whole capture via ?, and in_scope uses <= on what JVMTI defines as a half-open range, so boundary slots get read out of scope.

@buongarzoni Curious what you think.

waltjones (approved):
👍 All looks good.

